PR #46, opened 2026-02-18
Core service db and api: 403/404 error handling for 2 already made GET endpoints added and a small file structure fix 

greyear:


backend/services/api-gateway/src/middleware/auth.js
```diff
@@ -0,0 +1,64 @@
...
+    // Extract token from Authorization header (format: "Bearer <token>", standard RFC 6750 (OAuth 2.0))
+    const authHeader = req.headers.authorization;
+    
+    if (!authHeader || !authHeader.startsWith("Bearer ")) {
+      // No token provided - continue as guest
+      req.userId = null;
+      return next();
+    }
+
+    const token = authHeader.substring(7); // Remove "Bearer " prefix
```
FPyMEHTAPIU: This can be unified for all types. Since we might have something else than "Bearer". This also protects in case if someone put something else after the token.
```suggestion
    const token = authHeader.split(" ")[1]; // Remove "Bearer " prefix
```

backend/services/api-gateway/src/middleware/auth.js
```diff
@@ -0,0 +1,64 @@
...
+    // Validate token with auth-service
+    // Note: Auth-service validates the token once here, then we pass userId via headers
+    // to downstream microservices instead of making them validate the token again
+    try {
+      const response = await fetch(`${AUTH_SERVICE_URL}/auth/validate`, {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({ token }),
```
FPyMEHTAPIU: In body we usually provide the data like `username`, `password`, `name`. It's a good practice to keep the token in `headers.autorization`. And no need to use JSON.stringify.

FPyMEHTAPIU (approved):

PR #52, opened 2026-02-26
El testing

el-kittiwake:
Should only have 2 files with alterations. auth_schema.ts and auth_db_operations.ts

Basic auth_db functions should be in place now.

post:

    /login
    /register

get, delete, put:

    /users/:username

get:

    /users

Password hashing is done using bcrypt. JWT is generated on login and res.attach() to the res document.

Lots to do and improve, but at least this all works in unit testing.

Main concerns I have now: correctly using TS and how we are working with JWT.

### DIFF
--- a/backend/services/auth-service/package.json
+++ b/backend/services/auth-service/package.json
@@ -14,7 +14,8 @@
     "bcrypt": "^6.0.0",
     "express": "^5.2.1",
     "jsonwebtoken": "^9.0.3",
-    "mongoose": "^9.2.3"
+    "mongoose": "^9.2.3",
+    "zod": "^4.3.6"
   },
   "devDependencies": {
     "@types/bcrypt": "^6.0.0",
--- a/backend/services/auth-service/src/authGetSet.ts
+++ b/backend/services/auth-service/src/authGetSet.ts
@@ -17,7 +17,7 @@ export const authGetSet = Router();
 		1. Attempt to hash new password
 		2. findOneAndDelete() to remove matching record
 		4. Return relevant code
-*/
+*
 authGetSet.delete('/users/:username', async (req, res) =>
 {
 	try {
@@ -29,9 +29,10 @@ authGetSet.delete('/users/:username', async (req, res) =>
 		return res.json({ message: 'User deleted' });
 	
 	} catch (error) {
-		res.status(400).json({ error: error.message });
+		res.status(500).json({ error: error.message });
 	}
 });
+*/
 
 //Fetch single user record
 //Return all but passwordHash
@@ -47,7 +48,7 @@ authGetSet.get('/users/:username', async (req, res) =>
 		return res.json(userDocument);
 
 	} catch (error) {
-		res.status(400).json({ error: error.message });
+		res.status(500).json({ error: error.message });
 	}
 });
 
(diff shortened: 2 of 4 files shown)


backend/infrastructure/mongo/auth/init-auth.js
FPyMEHTAPIU: This file should be in TypeScript.

backend/infrastructure/mongo/db/auth_db_operations.ts
```diff
@@ -0,0 +1,245 @@
...
+//Connection port
+const port = 3000;
+
+// Parse incoming JSON requests automatically
+//https://www.geeksforgeeks.org/node-js/getting-started-with-express-js/
+app.use(express.json());
+
+// Connect to MongoDB
+// https://mongoosejs.com/docs/connections.html
+mongoose.connect('mongodb://127.0.0.1:27017/auth_db').then(() =>
```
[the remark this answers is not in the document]
FPyMEHTAPIU: We don't want to store any links in the code because sometimes they may contain sensitive data. We can only use some endpoints like `register/` or `login/`.

backend/infrastructure/mongo/db/auth_db_operations.ts
```diff
@@ -0,0 +1,245 @@
...
+{
+    console.log('Connected to MongoDB');
+    // Start the server after the database connection is established
+    app.listen(port, () => {
+    	console.log(`Server is running on http://localhost:${port}`);
+    });
+}).catch((err) => {
+	console.log('Error connecting to MongoDB:', err);
```
FPyMEHTAPIU: Should be `console.error` instead.

backend/infrastructure/mongo/db/auth_db_operations.ts
```diff
@@ -0,0 +1,245 @@
...
+{
+	try {
+		const document = await userModel.findOne(
+			{ $or: [
+					{email: req.body.username},
+					{username: req.body.username}
+				 ]
+			} );
+
+		if (!document)
```
FPyMEHTAPIU: I assume the `document` is a registered user. So can we call this variable `existingUser`?

backend/infrastructure/mongo/db/auth_db_operations.ts
```diff
@@ -0,0 +1,245 @@
...
+		{
+			const hashedPassword = await hashPassword(req.body.password);
+			if (hashedPassword)
```
FPyMEHTAPIU: Same early return would be cleaner.

backend/infrastructure/mongo/db/auth_db_operations.ts
```diff
@@ -0,0 +1,245 @@
...
+
+		if (!document)
+			res.status(404).json({error: 'Records not found'});
+		else
+			res.json(document);
+
+	} catch (error) {
+		res.status(400).json({ error: error.message });
+	}
+});
```
FPyMEHTAPIU: Same here. Move to another file.
`document` name is confusing to me. I don't understand what does it mean. The name should provide the purpose of the variable. As well as functions.
Check what do you return in `document`. Do we want to return everything from there?

FPyMEHTAPIU (changes requested):
Hey! That's a good start. The logic of the endpoints is mostly correct.

I'd recommend to get familiarized with [ESM](https://www.w3schools.com/nodejs/nodejs_modules_esm.asp). 

Don't take any comments personally, it's just part of learning. I spent months to gather the knowledge I'm trying to share here and I did similar mistakes. This is fine! 
If you would like me to elaborate one or another comment - feel free to ask here under one my comments or just write me a message in Discord. 🙌

backend/infrastructure/mongo/db/auth_db_operations.ts
el-kittiwake: Yeah, document is the record (document) from the auth DB, so indeed a registered user.
It is null if it is not found when using some search methods.

backend/infrastructure/mongo/db/auth_db_operations.ts
el-kittiwake: document is the literal result of the findOne() method. Either null if not found or the first document in the DB that matches the search criterion.

The contents of document should not be revealed to the outside world. Contents are username, email, passwordhash.

PR #92, opened 2026-03-06
Moving helpers to own file. Added header token validation function. V…

el-kittiwake:
Right-o.

I have attempted to add JWT validation to some degree.

1. endpoint: /auth/validate Takes no parameters but requires the authoriazation header to be in the format of: Authorization: <type> <token> returns username if valid and "Invalid token" all other times.

2. delete and change password endpoints now validate the JWT with the same requirements as above.
change password now also requires the old and new password as req.body.

Trying to get stuff done in a hurry so it can be turned around quickly.

Have at it reviewers!

### DIFF
--- a/backend/services/auth-service/src/authGetSet.ts
+++ b/backend/services/auth-service/src/authGetSet.ts
@@ -1,11 +1,13 @@
-import { Router, Request, Response } from 'express';
+import { Router, Request, Response, NextFunction } from 'express';
 
 // Import of project modules
 //Location of userModel may or may not change later.
 import { userModel } from './auth_schema.ts'; 
 import * as help from './authHelpers.ts';
+import { JwtPayload } from 'jsonwebtoken';
 
 export const authGetSet = Router();
+authGetSet.use(help.errorHandler);
 
 /*
 	Delete user. /users/:username endpoint
@@ -14,7 +16,7 @@ export const authGetSet = Router();
 		2. Return relevant code
 	As of now does not require the current password.
 */
-authGetSet.delete('/users/:username', help.compareJWT,  async (req: Request, res: Response) =>
+authGetSet.delete('/users/:username', help.compareJWT,  async (req: Request, res: Response, next: NextFunction) =>
 {
 	try {
 		const username = req.params.username;
@@ -27,13 +29,13 @@ authGetSet.delete('/users/:username', help.compareJWT,  async (req: Request, res
 		return res.json({ message: 'User deleted' });
 	
 	} catch (error) {
-		res.status(500).json({ error: error.message });
+		next(error);
 	}
 });
 
 //Fetch single user record
 //Return all but passwordHash
-authGetSet.get('/users/:username', async (req: Request, res: Response) =>
+authGetSet.get('/users/:username', async (req: Request, res: Response, next: NextFunction) =>
 {
 	try {
 		const username = req.params.username;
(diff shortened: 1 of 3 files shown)


backend/services/auth-service/src/authGetSet.ts
```diff
@@ -71,35 +73,66 @@ authGetSet.get('/users', async (req, res) =>
...
-		3. If good, create JWT and return
-		4. Return relevant code
-*
+		6. Return relevant code
+*/
 authGetSet.put('/users/:username', async (req, res) =>
 {
 	try {
-		if (!validatePassword(req.body.password))
+		const {newpassword, password} = req.body;
```
FPyMEHTAPIU: We should stick to the naming conventions, we use camelCase in the project. 🙌  
```suggestion
		const {newPassword, password} = req.body;
```

backend/services/auth-service/src/authGetSet.ts
```diff
@@ -71,35 +73,66 @@ authGetSet.get('/users', async (req, res) =>
...
+			return res.status(401).json({ error: 'Password mismatch' });
 
-		const hashedPassword = await hashPassword(req.body.password);
+		const hashedPassword = await help.hashPassword(newpassword);
 		if (!hashedPassword)
 			return res.status(500).json({error: 'Hashing failed'});
 
-		const userDocument = await userModel.findOneAndUpdate(
-			{username: req.params.username},
+		const toReplaceDoc = await userModel.findOneAndUpdate(
```
FPyMEHTAPIU: It can be called `updatedUser`.

backend/services/auth-service/src/authHelpers.ts
```diff
@@ -0,0 +1,116 @@
...
+	try {
+		const authHeaders = req.headers.authorization;
+		if (!authHeaders)
+			return null;
+
+		const testToken = authHeaders.split(" ");
+		const JWTSecret = process.env.JWTSecret || "placeholder";
+		const decoded = jwt.verify(testToken[1], JWTSecret);
+		console.log(decoded);
+		return decoded.username;
```
FPyMEHTAPIU: Double-check that you want to return a username.

backend/services/auth-service/src/authGetSet.ts
```diff
@@ -71,35 +73,66 @@ authGetSet.get('/users', async (req, res) =>
...
+*/
 authGetSet.put('/users/:username', async (req, res) =>
 {
 	try {
-		if (!validatePassword(req.body.password))
+		const {newpassword, password} = req.body;
+		const username = req.params.username;
+
+		if (!help.compareJWT(req))
+			return res.status(401).json({ error: "Invalid token" });
```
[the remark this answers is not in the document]
FPyMEHTAPIU: You can read about middlewares for express js more. One of my suggestions: https://www.w3schools.com/nodejs/nodejs_middleware.asp

backend/services/auth-service/src/authGetSet.ts
```diff
@@ -71,35 +70,69 @@ authGetSet.get('/users', async (req, res) =>
...
+		const username = decodedToken.username;
+		const userDocument = await userModel.findOne({username});
+		if (!userDocument)
+			return res.status(401).json({ error: "Invalid token" });
+
+		const userID = userDocument.get('_id');
+		return res.status(200).json({ _id: userID });
+	} catch (error) {
+		console.error(error);
+		return res.status(500).json({ error: error.message });
```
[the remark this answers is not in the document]
el-kittiwake: I am not getting this error in my vscode. I am getting others, but not this nor the ones that are similar.

PR #120, opened 2026-03-19
Auth user id

el-kittiwake:
Unique user ID should now function OK.
/auth/validate should now return id and not _id.


el-kittiwake:
This should cover everything. Dockerfile not changed just yet though.

FPyMEHTAPIU (approved):
Looks nice! Let's change the Docker later when we will be testing all the services.

PR #121, opened 2026-03-20
Create SearchField component

### LINKED ISSUE #31: Create search component

SeelaSalorinta:
Features:

- Form-based search (submit via Enter)
- Internal input state
- Clear (❌) button resets input and keeps focus
- Focus state styling (border highlight)
- Two modes: always-open (default) & collapsible (icon → input with back arrow)
- Escape: blurs input (always-open), closes search (collapsible)
- Collapsible mode uses Enter for submit (no icon to avoid crowded mobile layout)

close #31


SeelaSalorinta:
Changed the FormEvent to SyntheticEvent, as i understood that works better here than using the browser's SubmitEvent

PR #123, opened 2026-03-20
Header

### LINKED ISSUE #29: Create header component

FPyMEHTAPIU:
TODO: 
- change IconButton to have transparent variant.

Won't change Language selector here (or maybe ever), since I've already made quite many changes to other files.

closes #29

### DIFF
--- a/frontend/app/assets/styles/iconButton.css
+++ b/frontend/app/assets/styles/iconButton.css
@@ -9,9 +9,9 @@
 	justify-content: center;
 	cursor: pointer;
 	transition: filter 0.2s ease;
-	background-color: var(--bg-main);
+	background-color: var(--surface-primary);
 	padding: 0;
-	border: 1px solid var(--borders-action);
+	border: 1px solid var(--borders-input);
 }
 
 .icon-button svg {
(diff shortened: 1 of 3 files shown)


SeelaSalorinta (commented):
tested locally and works nicely overall 🙂
header structure and responsiveness feel really good! i like the layout 💫

also thanks for adding comments to explain some of the changes, that was really helpful!

i noticed UI inconsistencies and wanted to check with you:

- the search felt a bit jumpy when typing, not sure if intentional as it seems to expand slightly to the left

header elements feel a bit visually inconsistent right now:

- search uses a different background and border style than the other controls
- icon buttons (bell / menu / close in header) and language buttons (outside header) use bg-main with black text/icons, while header elements use surface-primary and green text
- borders are also mixed in header: some controls have no border, some green borders, and some lighter ones, so it feels a bit inconsistent
- desktop shows only "sign up" while mobile menu has "log in / sign up"

the header could maybe use one consistent theme for all controls, like same background, borders and icon or text color
for example we could use borders-default for all and surface-primary as background, and then text-primary / icons-dark / text-on-accent

FPyMEHTAPIU:
Good catches, thank you for such a detailed review! 🫶 
I'll change the styles and keep the buttons consistent. For the login button I haven't decided yet how to handle it better (Log in/Sign Up seems a bit long). And having two separate buttons might seem too crowdy.
For the search bar - I decided not to touch it in this PR, since it's a separate component. But as I mentioned, it'd be nice to have expanding animation for it and fix this jump.

PR #131, opened 2026-03-24
Early cookie and session things. 

el-kittiwake:
Not very useful as of now, but it's a start.

PR so we have something to talk over on Wednesday. I feel like I'm banging my head against problems working with typescript. Everything works as I suppose it should when unit tested, but falls apart as it gets integrated.

### DIFF
--- a/backend/services/auth-service/src/auth_db_operations.ts
+++ b/backend/services/auth-service/src/auth_db_operations.ts
@@ -166,6 +166,7 @@ authRouter.post(
 	https://developers.google.com/identity/gsi/web/guides/verify-google-id-token
 	https://www.w3tutorials.net/blog/google-sign-in-backend-verification/
 */
+/*
 authRouter.post(
 	"/google",
 	async (req: Request, res: Response, next: NextFunction) => {
@@ -190,11 +191,9 @@ authRouter.post(
 			const googleID = payload.sub;
 			const { email, name } = payload;
 
-			/*
-			Repetiton here, which can be sorted out later.
-			Google accounts will not require a passwordHash, so just using "empty"
-			Not sure how correct any of this is, but making a start.
-			*/
+			//Repetiton here, which can be sorted out later.
+			//Google accounts will not require a passwordHash, so just using "empty"
+			//Not sure how correct any of this is, but making a start.
 			const userDocument = await userModel.findOne({ googleID });
 			if (!userDocument) {
 				const currentCount = await help.makeID();
@@ -230,3 +229,4 @@ authRouter.post(
 		}
 	},
 );
+*/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,8 +14,7 @@ services:
     ports:
       - "27017:27017"
 
-    volumes:
-      - ./backend/infrastructure/mongo/auth/init-auth.js:/docker-entrypoint-initdb.d/init-auth.js:ro
+	  #We had volumes here, but we removed it.
 
     networks:
       - transcendence-network
--- a/backend/services/auth-service/src/authGetSet.ts
+++ b/backend/services/auth-service/src/authGetSet.ts
@@ -134,6 +134,7 @@ authGetSet.post(
 
 // /auth/validate/google endpoint to specifically validate a JWT of a google account
 // within the header.
+/*
 authGetSet.post(
 	"/validate/google",
 	async (req: Request, res: Response, next: NextFunction) => {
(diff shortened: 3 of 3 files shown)


backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -5,25 +5,47 @@
...
+		user: string;
+	}
+}
+
 export const authRouter = Router();
-authRouter.use(help.errorHandler);
+// Middleware setup
+authRouter.use(
+	session({
+		secret: "placeholder",
```
greyear: Should there be placeholder? 
Should we create separate environment variable and use it instead?

backend/services/api-gateway/src/routes/auth.routes.ts
```diff
@@ -0,0 +1,196 @@
...
+		// Forward Authorization header if present
+		if (req.headers.authorization) {
+			headers.authorization = req.headers.authorization;
+		}
+		const response = await fetch(`${AUTH_SERVICE_URL}/validate`, {
+			method: "POST",
+			headers,
+			body: JSON.stringify(req.body),
+		});
+		console.log(response);
```
greyear: That would be [object Response] from fetch every time. What is the point?

docker-compose.yml
```diff
@@ -17,6 +17,9 @@ services:
     volumes:
       - ./backend/infrastructure/mongo/auth/init-auth.js:/docker-entrypoint-initdb.d/init-auth.js:ro
```
greyear: I tried to clone repo and build command complained about this file's permissions (file is created on build)

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -130,7 +152,11 @@ authRouter.post(
 				return;
 			}
 
+			// Send success response
 			const JWToken = help.generateToken(userDocument.get("_id"), username);
+			req.session.user = username;
+			res.cookie("sessionId", req.sessionID);
+			res.cookie("JWToken", JWToken);
```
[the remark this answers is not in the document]
el-kittiwake: OK, changed to those defaults for now.

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -5,25 +5,47 @@
...
-authRouter.use(help.errorHandler);
+// Middleware setup
+authRouter.use(
+	session({
+		secret: "placeholder",
+		resave: false,
+		saveUninitialized: true,
+	}),
+);
+authRouter.use(cookieParser());
```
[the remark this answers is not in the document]
el-kittiwake: It is already declared in api-gateway/app.ts, so I just remove it from here.

backend/services/api-gateway/src/routes/auth.routes.ts
el-kittiwake: Every instance of console.log() (at least in my part) is for debugging. This seems to be functioning as it should, so I will remove them.

el-kittiwake:
Made changes. Manual testing still passing. Google auth can't really be tested until I have tokens to test with, which I do not create. Google doesn't seem to provide test users to do unit testing with.

I cannot reproduce make problems, and two fresh clones built fine.

Anyway, even if it isn't all fine, many things should be.

docker-compose.yml
greyear: I tried just now and got it again:
Creating auth-mongo            ... error

Creating core-postgres         ... done
Creating notification-postgres ... done
frastructure/mongo/auth/init-auth.js: operation not permitted
Creating notification-service  ... done
Creating core-service          ... done

ERROR: for auth-db  Cannot start service auth-db: error while creating mount source path '/vagrant/github/hive_projects/virgin/backend/infrastructure/mongo/auth/init-auth.js': chown /vagrant/github/hive_projects/virgin/backend/infrastructure/mongo/auth/init-auth.js: operation not permitted
ERROR: Encountered errors while bringing up the project.
make: *** [Makefile:53: re] Error 1

greyear:
Thanks, Eric. Your fixes look mostly fine, the only 2 things are 1) build failure with incorrect permissions of the created file 2) testing of google endpoints which probably can fail for the reason I mentioned above, but I understand that it's more handy to check and fix when there'll be a possibility to do it together with frontend

docker-compose.yml
el-kittiwake: Interesting, I'll look into that. Did AI offer anything useful?

PR #132, opened 2026-03-24
Post recipes endpoints

### LINKED ISSUE #21: Implement recipe /create endpoint
This must take all the required fields (see Figma) and add the recipe to the database. Steps might be represented as array of TEXT or just a string with \n delimeter (might not be the best solution).

### LINKED ISSUE #122: Implement POST   /recipes/{id}/publish
Recipe status changed to published (waits moderation)

greyear:
Added two endpoints realization: 
POST   /recipes (closes #21)
POST   /recipes/{id}/publish (closes #122)
Description can be found on google sheets: transcendence/endpoints

### DIFF
--- a/backend/services/core-service/src/services/recipes.service.ts
+++ b/backend/services/core-service/src/services/recipes.service.ts
@@ -44,6 +44,11 @@ const recipeVisibilityRowSchema = z.object({
 	status: recipeStatusSchema,
 });
 
+const rowIdSchema = z
+	.object({ id: z.union([z.number(), z.string()]) })
+	.transform((r) => r.id)
+	.catch("unknown");
+
 const getRecipeWithIngredientsQuery = `
 	SELECT
 		r.id,
@@ -105,15 +110,7 @@ const parseRecipeRows = <T>(
 		if (validation.success) {
 			acc.push(validation.data);
 		} else {
-			const maybeRow =
-				typeof row === "object" && row !== null
-					? (row as { id?: unknown })
-					: null;
-
-			const rowId =
-				typeof maybeRow?.id === "number" || typeof maybeRow?.id === "string"
-					? maybeRow.id
-					: "unknown";
+			const rowId = rowIdSchema.parse(row);
 
 			console.error(
 				`Skipping invalid ${rowLabel} ID ${rowId}:`,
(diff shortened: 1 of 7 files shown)


backend/services/core-service/src/services/recipes.service.ts
```diff
@@ -34,9 +105,14 @@ const parseRecipeRows = <T>(
...
+				typeof row === "object" && row !== null
+					? (row as { id?: unknown })
+					: null;
+
 			const rowId =
-				typeof row.id === "number" || typeof row.id === "string"
-					? row.id
+				typeof maybeRow?.id === "number" || typeof maybeRow?.id === "string"
+					? maybeRow.id
 					: "unknown";
```
FPyMEHTAPIU: Consider replacing it by Zod and always avoid "as" casting:

```
	const rowId = z
        .object({ id: z.union([z.number(), z.string()]) })
        .transform((r) => r.id)
        .catch("unknown")
        .parse(row);

    console.error(
        `Skipping invalid ${rowLabel} ID ${rowId}:`,
        z.prettifyError(validation.error),
    );
```
or 

```
const rowIdSchema = z
    .object({ id: z.union([z.number(), z.string()]) })
    .transform((r) => r.id)
    .catch("unknown");

// usage:
const rowId = rowIdSchema.parse(row);

```

backend/services/core-service/src/services/recipes.service.ts
```diff
@@ -151,6 +254,140 @@ export const getMyRecipes = async (
...
+          VALUES ($1, $2, $3, $4)
+        `,
+				[
+					recipeId,
+					ingredient.ingredient_id,
+					ingredient.amount,
+					ingredient.unit,
+				],
+			);
+		}
```
FPyMEHTAPIU: This might be a performance bottleneck since you're doing many queries in a for loop. Consider something like this:
```suggestion
const ingredientIds: number[] = [];
const amounts: number[] = [];
const units: string[] = [];

for (const { ingredient_id, amount, unit } of input.ingredients) {
    ingredientIds.push(ingredient_id);
    amounts.push(amount);
    units.push(unit);
}

await client.query(
    `
    INSERT INTO recipe_ingredients (recipe_id, ingredient_id, amount, unit)
    SELECT $1, unnest($2::int[]), unnest($3::numeric[]), unnest($4::text[])
    `,
    [recipeId, ingredientIds, amounts, units],
);
```

backend/services/core-service/src/services/recipes.service.ts
```diff
@@ -151,6 +254,140 @@ export const getMyRecipes = async (
...
+
+		for (const categoryId of input.category_ids) {
+			await client.query(
+				`
+          INSERT INTO recipe_category_map (recipe_id, category_id)
+          VALUES ($1, $2)
+        `,
+				[recipeId, categoryId],
+			);
+		}
```
FPyMEHTAPIU: Same unnest approach here:

```suggestion
		if (input.category_ids.length > 0) {
			await client.query(
				`
				INSERT INTO recipe_category_map (recipe_id, category_id)
				SELECT $1, unnest($2::int[])
				`,
				[recipeId, input.category_ids],
			);
		}
```

backend/services/core-service/src/validation/schemas.ts
```diff
@@ -39,7 +39,61 @@ const recipeIdSchema = intIdSchema;
...
+});
+
+const createRecipeCategoryIdSchema = intIdSchema;
+
+const createRecipeInputSchema = z.object({
+	title: z.string().trim().min(1).max(256),
+	description: z.string().max(5000).optional().nullable(),
+	instructions: z.array(z.string().trim().min(1)).min(1),
+	servings: intIdSchema.optional().default(1),
+	spiciness: z.coerce.number().int().min(0).max(3).optional().default(0),
```
FPyMEHTAPIU: And these are not optional.

backend/services/core-service/src/validation/schemas.ts
```diff
@@ -71,6 +125,36 @@ export const validateUserId = (id: unknown): ValidationResult => {
 	};
 };
 
+type CreateRecipeValidationResult =
+	| { valid: true; value: CreateRecipeInput }
```
FPyMEHTAPIU: Since now we use similar structure we can use templates.
```
type ValidationResult<T> =
	| { valid: true; value: T }
	| { valid: false; error: string };
```

test-endpoints.sh
```diff
@@ -66,6 +125,29 @@ check_endpoint "/recipes/.1" "400" "GET /recipes/.1 (starts with dot)"
 check_endpoint "/recipes/40a" "400" "GET /recipes/40a (alphanumeric)"
 check_endpoint "/users/abc/recipes" "400" "GET /users/abc/recipes (invalid user ID)"
 
+echo ""
+echo "POST Endpoints:"
+check_post_endpoint "/recipes" '{"title":"Smoke Recipe","description":"Created by smoke test","prep_time_minutes":10,"cook_time_minutes":20,"servings":2,"difficulty":"easy","ingredients":[{"ingredient_name":"Water","quantity":"1 cup"}]}' "401" "POST /recipes (no token -> 401)"
```
FPyMEHTAPIU: does `"ingredients":[{"ingredient_name":"Water","quantity":"1 cup"}]` match the recipe schema? 
There were ingredient_id, name, amount and unit.

PR #140, opened 2026-03-28
add floating labels to InputField

### LINKED ISSUE #138: Add floating labels to InputField

SeelaSalorinta:
added floating label support to InputField when only placeholder is provided

- label now floats on focus / value
- keeps normal label when label prop is used

close #138

### DIFF
--- a/frontend/app/assets/styles/inputField.css
+++ b/frontend/app/assets/styles/inputField.css
@@ -69,19 +69,16 @@
 
 .input-wrapper.floating .input-field {
 	height: 100%;
-	padding-top: 24px;
-	padding-bottom: 6px;
-	padding-right: 36px;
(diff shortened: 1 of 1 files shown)


frontend/app/assets/styles/inputField.css
```diff
@@ -62,16 +62,43 @@
...
+.floating-label {
 	position: absolute;
-	width: 1px;
-	height: 1px;
+	left: 14px;
+	top: 50%;
+	transform: translateY(-50%);
+	color: var(--text-secondary);
+	pointer-events: none;
+	background-color: none;
```
FPyMEHTAPIU: "none" is not valid option for the background-color:

```suggestion
	background-color: transparent;
```

PR #157, opened 2026-04-04
Sort button

### LINKED ISSUE #151: Implement sort button

FPyMEHTAPIU:
- Users can sort recipes by name (A→Z or Z→A) using a dropdown menu. 
- The selected sort is persisted in the URL as a query parameter, which resets to default when the page is refreshed or the default option is re-selected. 
- Changing the sort order automatically resets pagination to page 1. 
- Sort option labels are fully translated across all supported languages.

closes #151


SeelaSalorinta (dismissed):
great job! nice to see this implementation, looks really good visually and works smoothly 💯

Noticed these aren't translated yet:
- people page sort options
- aria-label="Sort by"

otherwise looks good!

SeelaSalorinta (dismissed):
nice refactor and still works great ! 🥇  aria-label="Sort by" is still not translated if you meant to add that too

PR #183, opened 2026-04-08
Added Google auth basics. 

el-kittiwake:
Tests manually fine with a token generated in the oauthplayground. AI tests work, testing those cases manually works. I feel this is a good start, but will need work and cleanup once problems are ironed out.

Also added check for valid username (3-20 chars, basic characters).

Google auth token will need creating using client ID that is pinned in the discord. Secret can be shared there if needed, or I can create a token if asked.


backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -166,12 +185,11 @@ authRouter.post(
 	https://developers.google.com/identity/gsi/web/guides/verify-google-id-token
 	https://www.w3tutorials.net/blog/google-sign-in-backend-verification/
 */
-/*
 authRouter.post(
 	"/google",
 	async (req: Request, res: Response, next: NextFunction) => {
 		try {
-			const CLIENT_ID = process.env.GOOGLE_CLIENT_ID;
+			const CLIENT_ID = process.env.GOOGLE_CLIENT_ID || "443643296362-p5t0avftu3eu6nf78p9pv6ot5adomors.apps.googleusercontent.com";
```
greyear: This might be potentially unsafe. 
- Hardcoding credentials can expose your app config publicly, especially if this is in a repo
- You might accidentally use the fallback in production if env vars are misconfigured
- Even though a Google Client ID isn’t as sensitive as a secret key, it still shouldn’t be hardcoded in backend logic

I would make it so if your env var is failing - just check for an error

const CLIENT_ID = process.env.GOOGLE_CLIENT_ID;

if (!CLIENT_ID) {
  throw new Error("Missing GOOGLE_CLIENT_ID in environment variables");
}

(Reviewed by DIma)

backend/services/auth-service/src/authHelpers.ts
```diff
@@ -1,7 +1,7 @@
 import bcrypt from "bcrypt";
 import type { NextFunction, Request, Response } from "express";
 import jwt, { type JwtPayload } from "jsonwebtoken";
-import z from "zod";
+import z, { refine } from "zod";
```
greyear: Looks like never used anywhere

(Reviewed by Dima)

greyear (changes requested):
Please consider these suggested changes. I would also ask to move google auth tests (and all .sh scripts in general) to the separate folder in order to make it more neat.

(reviewed by Dima)

greyear (approved):
I would maybe merge that PR, overall it looks good visually and well documented now. And then we need to test this together with frontend later on. 
Thanks, Eric.

(Dima)

PR #189, opened 2026-04-09
Recipe picture upload

greyear:
Now recipes have their own pics (one per recipe!)


el-kittiwake (approved):
Looks good to me! Thanks.

PR #190, opened 2026-04-10
Added login notification via api-gateway to core-service.

el-kittiwake:
Removed realname from authSchema, logic, tests and documentation. Added isActive field to authDB.
Added logic to check for inactivity and make inactive if deleted. This is to prevent someone creating a new user with an old username.


el-kittiwake:
In the auth-routes gateway code there are 2 commented fetch() to the core-service to account for logon. Should function with them commented.

backend/services/api-gateway/src/routes/auth.routes.ts
```diff
@@ -47,11 +48,37 @@ const postLoginHandler: RequestHandler = async (
...
+			const validateResponse = await fetch(`${AUTH_SERVICE_URL}/validate`, {
+				method: "POST",
+				headers: validateHeaders,
+			});
+
+			const validateData = (await validateResponse.json()) as { id: string };
+			const userId = validateData.id;
+
+			// Update user status to "online" in core service
+			// Endpoint and final logic will probably change.
```
greyear: (Comment from Dima)

My thoughts on this - we should implement heartbeat exchange logic instead. The reason is that it's much easier and doesn't bring extra complexity to the services logic. I will repeat my comment I've left in chat here once more so when @FPyMEHTAPIU will review this PR he could express his point of view.

"So, I think the best solution I found so far:

Frontend sends a POST /users/me/heartbeat every 30–60 seconds. Core-service sets status = 'online' and stores last_seen_at.

- Frontend sends heartbeat every 30–60s → core-service sets online + last_seen_at
- On logout → send /users/me/offline, then clear the cookie
- Token expires or tab closes → no more heartbeats → server considers user offline after timeout (backend worker should do those checks)
 
So basically we need to implement 2 endpoints in core service + FE to call them:
1) **POST /users/me/heartbeat** - called every 30s by the frontend and sets status = 'online' and last_seen_at = now()
2) **POST /users/me/offline** - called on logout, sets status = 'offline'"

el-kittiwake: I'm happy with whatever is decided. I just need to know what is expected from me :)

el-kittiwake:
Fixed merge conflicts in the web editor. Which was not a fun experience. Needs review again.

backend/services/api-gateway/src/routes/auth.routes.ts
```diff
@@ -47,11 +48,37 @@ const postLoginHandler: RequestHandler = async (
 			body: JSON.stringify(req.body),
 		});
 		const data = await response.json();
+
+		// Extract userId from validated token
+		if (response.status === 200 && (data as { token: string }).token) {
```
FPyMEHTAPIU: Let's not use type assertions "as" keyword. It might break later if token is undefined.
Create this type in the beginning of the file:
```
const tokenResponseSchema = z.object({ token: z.string() });
```
and move the code below to the postAuthRegister,

backend/services/api-gateway/src/routes/auth.routes.ts
FPyMEHTAPIU: No need for the second endpoint. I agree with the first one.

backend/services/api-gateway/src/routes/auth.routes.ts
```diff
@@ -47,11 +48,37 @@ const postLoginHandler: RequestHandler = async (
...
+				Authorization: `Bearer ${(data as { token: string }).token}`,
+			};
+
+			const validateResponse = await fetch(`${AUTH_SERVICE_URL}/validate`, {
+				method: "POST",
+				headers: validateHeaders,
+			});
+
+			const validateData = (await validateResponse.json()) as { id: string };
+			const userId = validateData.id;
```
FPyMEHTAPIU:
```suggestion
```

backend/services/api-gateway/src/routes/auth.routes.ts
FPyMEHTAPIU, line 35:
```suggestion
const response = await fetch(`${AUTH_SERVICE_URL}/register`, {
method: "POST",
headers: { "Content-Type": "application/json" },
body: JSON.stringify(req.body),
});
const data = await response.json();
if (response.status === 201 && tokenResponseSchema.safeParse(data).success) {
			const setCookieHeader = response.headers.get("set-cookie");
			if (setCookieHeader) {
				res.set("Set-Cookie", setCookieHeader);
			}
		}
```

backend/services/api-gateway/src/routes/auth.routes.ts
```diff
@@ -77,11 +104,37 @@ const postGoogleHandler: RequestHandler = async (
...
+
+			// Update user status to "online" in core service
+			// Endpoint and final logic will probably change.
+			/*await fetch(`${CORE_SERVICE_URL}/users/${userId}/status`, {
+				method: "PATCH",
+				headers: { "Content-Type": "application/json" },
+				body: JSON.stringify({ status: "online" }),
+			}); */
+		}
```
FPyMEHTAPIU: Seems like the same code as above. Remove it.

backend/services/api-gateway/src/routes/recipes.routes.ts
```diff
@@ -307,3 +307,5 @@ recipesRouter.delete("/:id", requireAuth, deleteRecipeHandler);
 
 recipesRouter.get("/", optionalAuth, getRecipesHandler);
 recipesRouter.post("/", requireAuth, createRecipeHandler);
+
+recipesRouter.use("/:id/rating", ratingsRouter);
```
FPyMEHTAPIU: It's already mentioned on line 301.
```suggestion
```

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -106,13 +109,24 @@ authRouter.post(
...
+			res.status(201).json({ username, email });
+			console.log("res.status:" ,res.status);
+
+			const loginRes = await fetch(`${AUTH_SERVICE_URL}/login`, {
+				method: "POST",
+				headers: { "Content-Type": "application/json" },
+				body: JSON.stringify(req.body),
+			});
+			console.log("request body: " ,req.body);
+			console.log("Login response: ", loginRes);
```
FPyMEHTAPIU: So this first sends a response and then sends `login/` fetch which doesn't change anything. This should happen before sending the response. Also let's not have `console.log`s in PRs 🙂

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -106,13 +109,24 @@ authRouter.post(
...
+
 		} catch (error) {
-			if ((error as any)?.code === 11000) {
+			if ((error as { code?: number })?.code === 11000) {
```
FPyMEHTAPIU: Let's use Zod for it. Create the type in the beginning of the file:
```
const mongoErrorSchema = z.object({ code: z.literal(11000) });
```
and then:

```suggestion
			if (mongoErrorSchema.safeParse(error).success) {
```

backend/services/auth-service/src/auth_db_operations.ts
FPyMEHTAPIU: This might also use Zod and might require a separate type for it:
```
const loginResponseSchema = z.object({
	token: z.string(),
	message: z.string(),
});
```

FPyMEHTAPIU: After fetch:
```
const loginParsed = loginResponseSchema.safeParse(await loginRes.json());
			const loginPayload = loginParsed.success ? loginParsed.data : {};

			const setCookie = loginRes.headers.get("set-cookie");
			if (setCookie) {
				res.set("Set-Cookie", setCookie);
			}
			res.status(201).json({ username, email, ...loginPayload });
```

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -171,6 +190,7 @@ authRouter.post(
 			const actualUsername = userDocument.get("username");
 			const JWToken = help.generateToken(
 				userDocument.get("_id"),
+				userDocument.get("id"),
 				actualUsername as string,
 				"mongo",
 			);
```
FPyMEHTAPIU: Let's also avoid type assertion:

```suggestion
const usernameResult = z.string().safeParse(userDocument.get("username"));
			if (!usernameResult.success) {
				res.status(500).json({ error: "Invalid username in database" });
				return;
			}
			const JWToken = help.generateToken(
				userDocument.get("_id"),
				userDocument.get("id"),
				usernameResult.data,
				"mongo",
			);
```

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -254,16 +270,30 @@ authRouter.post(
...
-				return;
+				
+				const loginRes = await fetch(`${AUTH_SERVICE_URL}/google`, {
+					method: "POST",
+					headers: { 
+						"Content-Type": "application/json",
+						"Authorization": `Bearer ${token}`
+					},
+				});
+				console.log("Login response: ", loginRes);
```
FPyMEHTAPIU: Same here:

```suggestion
				const loginRes = await fetch(`${AUTH_SERVICE_URL}/google`, {
					method: "POST",
					headers: {
						"Content-Type": "application/json",
						Authorization: `Bearer ${token}`,
					},
				});
				const loginParsed = loginResponseSchema.safeParse(await loginRes.json());
				const loginPayload = loginParsed.success ? loginParsed.data : {};

				const setCookie = loginRes.headers.get("set-cookie");
				if (setCookie) {
					res.set("Set-Cookie", setCookie);
				}
				res.status(201).json({ googleID, email, name, ...loginPayload });
```

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -281,7 +311,7 @@ authRouter.post(
 				return;
 			}
 		} catch (error) {
-			if ((error as any)?.code === 11000) {
+			if ((error as { code?: number })?.code === 11000) {
```
FPyMEHTAPIU: And use the type mentioned above.

backend/services/auth-service/src/auth_schema.ts
```diff
@@ -42,23 +49,24 @@ const authSchema = new mongoose.Schema({
...
 		type: String,
 		required: false,
 		unique: true,
 		sparse: true,
 	},
+	isActive: {
+		// realname is removed. isActive is added.
+		type: Boolean,
+		default: true,
+	},
```
FPyMEHTAPIU: Let's remove it since we won't have soft deletion.

backend/services/auth-service/src/auth_schema.ts
```diff
@@ -42,23 +49,24 @@ const authSchema = new mongoose.Schema({
 		type: String,
 		required: true,
 		unique: true,
+		sparse: true,
```
FPyMEHTAPIU: Let's remove this. There won't be any emails with NULL.

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -147,6 +161,11 @@ authRouter.post(
 				return;
 			}
 
+			if (!userDocument.get("isActive")) {
+				res.status(403).json({ error: "Account not valid." });
+				return;
+			}
```
FPyMEHTAPIU: Make sure you removed this and other similar pieces after fixing the schema.

backend/services/auth-service/src/authGetSet.ts
```diff
@@ -29,7 +29,11 @@ authGetSet.delete(
 
 			let userDocument = null;
 			if (req.decodedJWT?.type === "mongo") {
-				userDocument = await userModel.findOneAndDelete({ username });
+				userDocument = await userModel.findOneAndUpdate(
+					{ username },
+					{ email: null, isActive: false },
+					{ new: true },
+				);
```
FPyMEHTAPIU: can be reverted to 

```suggestion
				userDocument = await userModel.findOneAndDelete({ username });

```

backend/services/auth-service/src/authGetSet.ts
FPyMEHTAPIU: We can also get rid of this type assertions if we want to make it cleaner and more robust 🙂 .
See comments in authHelpers.ts
```suggestion
			const decodedToken = help.fetchDecodeToken(req);
```

backend/services/auth-service/src/authHelpers.ts
FPyMEHTAPIU, line 1: We can add a zod type for JWT payload:
```
export const jwtPayloadSchema = z.object({
	sub: z.string(),
	userId: z.number(),
	username: z.string(),
	type: z.string(),
});
export type AppJwtPayload = z.infer<typeof jwtPayloadSchema>;
```

FPyMEHTAPIU: And we can use this zod parsing here:
```suggestion
export const fetchDecodeToken = (req: Request): AppJwtPayload | null => {
	try {
		const tokenHeader = sequenceHeader(req);
		if (!tokenHeader) {
			throw Error("Given header does not contain a parseable token");
		}

		const decoded = decodeToken(tokenHeader);
		if (!decoded) {
				throw Error("Whatever was given was not a token");
		}

		const parsed = jwtPayloadSchema.safeParse(decoded);
		if (!parsed.success) {
			throw Error("Token payload does not match expected shape");
		}

		return parsed.data;
	} catch (error) {
```

FPyMEHTAPIU (changes requested):
Hey, that was a huge work done! 💪 
Mostly my comments about the type assertions which are dangerous, so I'd like to replace them.
Let's not have a status here for now. 🙌

backend/services/api-gateway/src/routes/auth.routes.ts
el-kittiwake: This is quite hard to follow. I mean the whole thing. Hopefully I don't make any mistakes with fixes.

Adding
const tokenResponseSchema = z.object({ token: z.string() });
to the top of auth.routes.ts

backend/services/api-gateway/src/routes/auth.routes.ts
el-kittiwake: All of this is removed.

So we are not sending notification of login to the core-service. Status is being handled in another way?

backend/services/api-gateway/src/routes/auth.routes.ts
el-kittiwake: So this is using zod to check that the token is a string? safeParse doesn't throw if there is some problem?

el-kittiwake:
Quick tests, everything seems fine.

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -112,21 +119,23 @@ authRouter.post(
...
-			console.log("Login response: ", loginRes);
+
+			const loginParsed = loginResponseSchema.safeParse(await loginRes.json());
+			const loginPayload = loginParsed.success ? loginParsed.data : {};
+
+			const setCookie = loginRes.headers.get("set-cookie");
+			if (setCookie) {
+				res.set("Set-Cookie", setCookie);
+			}
+			res.status(201).json({ username, email, id: currentCount, loginPayload });
```
FPyMEHTAPIU: loginPayload should be spread. Without the spread, the token was nested at data.loginPayload.token instead of the expected data.token.
```suggestion
            res.status(201).json({ username, email, id: currentCount, ...loginPayload });
```

backend/services/auth-service/src/auth_db_operations.ts
```diff
@@ -273,24 +281,23 @@ authRouter.post(
...
-					res.status(403).json({ error: "Account not valid." });
-					return;
+				const loginParsed = loginResponseSchema.safeParse(await loginRes.json());
+				const loginPayload = loginParsed.success ? loginParsed.data : {};
+				const setCookie = loginRes.headers.get("set-cookie");
+				if (setCookie) {
+					res.set("Set-Cookie", setCookie);
 				}
 
+				res.status(201).json({ googleID, email, name, id: currentCount, loginPayload });
```
FPyMEHTAPIU: Same here.

backend/services/auth-service/src/auth_schema.ts
FPyMEHTAPIU: This comment is stale now.

backend/services/auth-service/src/authGetSet.ts
FPyMEHTAPIU: unused import now

backend/services/auth-service/src/authHelpers.ts
FPyMEHTAPIU: And unused { type JwtPayload } here

FPyMEHTAPIU (commented):
Very good! 🙌 
Only a couple minor things left and then it's ready to merge.

backend/services/auth-service/src/auth_db_operations.ts
el-kittiwake: Oh, new thing to me.

FPyMEHTAPIU (approved):

PR #193, opened 2026-04-11
add initial auth popup for login and account creation

### LINKED ISSUE #60: Create register (sign up) popup (SEE DESCRIPTION)
In Figma it looks like a separate page, but I think it's also possible to have it as a popup. Not sure what solution we should stick to.

### LINKED ISSUE #61: Create login page

SeelaSalorinta:
What’s included:
- auth popup opened from header and footer Sign in -button
- login and signup shares the same form inside the popup
- login connected to /auth/login
- signup connected to /auth/register
- signup validation for username and password based on backend rules
- inputField shows hint only when focused now, not all the time like before
- local testing required two backend-side changes which i pushed here after discussing with Eric:
       - realname changed to optional in auth-service schema
       - credentials: true added in API gateway CORS config

Notes:
- Google login is still placeholder UI only and is not wired yet
- this PR focuses only on the initial auth popup setup (signed-in UI changes after login, such as replacing the Sign In button in header/footer, are intentionally left for a follow-up task to keep this PR smaller)
       
 close #61 
 close #60


frontend/app/components/auth/AuthModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+		window.addEventListener("keydown", handleEscape);
+		return () => window.removeEventListener("keydown", handleEscape);
+	}, [isOpen, onClose]);
+
+	if (!isOpen) {
+		return null;
+	}
+
+	return (
+		<div className="auth-modal-backdrop" role="presentation">
```
greyear, line 112: Please investigate this possible fix:

When the auth modal is open, a keyboard user pressing Tab can move focus to elements behind the backdrop — buttons in the header, footer links, etc. They're interacting with a page they can't see. A focus trap keeps Tab cycling only through elements inside the modal.

Scroll lock is the same idea for mouse/touch users. Without overflow: hidden on the body, you can scroll the background page while the modal is on top of it, which feels broken.

Both are standard expectations for any modal dialog, and since the component already declares aria-modal="true", assistive tech assumes these behaviours are in place.

So basically my suggestion is to add **Focus trap** and **scroll lock** for this popup to prevent malicious actions on the background.

greyear (changes requested):
Just one comment on modal window

(Dima)

SeelaSalorinta:
Thank you ! Should be fixed now!

greyear (approved):
Accepting

(Dima)

Issue #199, opened 2026-04-12
Delete the review


greyear:
<google-sheets-html-origin><style type="text/css"></style>
18-1 | Delete the review | User | User is authenticated, user owns review | recipe_reviews | Review is deleted | - user is not authenticated (401) - review not found (404) - user has no permission (403) | DELETE   /reviews/{id}
-- | -- | -- | -- | -- | -- | -- | --

PR #200, opened 2026-04-12
Reviews update/delete endpoints

### LINKED ISSUE #198: Update the review
<google-sheets-html-origin><style type="text/css"></style>
17-1 | Update the review | User | User is authenticated, user owns review | recipe_reviews | Review is updated | - user is not authenticated (401) - review not found (404) -user has no permission (403) | PUT   /reviews/{id}
-- | -- | -- | -- | -- | -- | -- | --

greyear:
Closes #198 #199 

Implemented:

PUT /recipes/{id}/reviews/{reviewId}
DELETE /recipes/{id}/reviews/{reviewId}


backend/services/core-service/src/services/recipes.service.ts
```diff
@@ -942,6 +951,130 @@ export const getRecipeReviews = async (
...
+				u.username,
+				u.avatar,
+				rr.body,
+				rr.created_at,
+				rr.updated_at
+			FROM recipe_reviews rr
+			LEFT JOIN users u ON u.id = rr.author_id
+			WHERE rr.id = $1
+		`,
+			[updateResult.rows[0].id],
```
el-kittiwake: After the UPDATE, a separate SELECT JOIN is made to fetch user data, but reviewId is already known — so updateResult.rows[0].id is unnecessary. Pass reviewId directly to the second query.

backend/services/api-gateway/src/routes/recipes.routes.ts
```diff
@@ -175,6 +175,53 @@ const getRecipeReviewsHandler: RequestHandler = async (req, res, _next) => {
...
+			res.status(504).json({ error: "Gateway Timeout" });
+			return;
+		}
+
+		console.error("Error proxying to core-service:", error);
+		res.status(500).json({ error: "Failed to update review" });
+	}
+};
+
+const deleteReviewHandler: RequestHandler = async (req, res, _next) => {
```
el-kittiwake, line 202: deleteReviewHandler returns IDs from request params, not from DB
The response builds data from reviewIdValidation.value / recipeIdValidation.value instead of actual DB output. This is inconsistent with updateReviewHandler, which returns a full DB object. Consider having deleteReview return updated_at from the DB so the client knows the actual deletion timestamp.

backend/services/core-service/src/routes/recipes.routes.ts
```diff
@@ -693,6 +696,153 @@ const getRecipeReviewsHandler = async (
...
+		}
+
+		const recipeIdValidation = validateRecipeId(req.params.id);
+		if (!recipeIdValidation.valid) {
+			const error: CustomError = new Error(recipeIdValidation.error);
+			error.statusCode = 400;
+			throw error;
+		}
+
+		const reviewIdValidation = validateRecipeId(req.params.reviewId);
```
el-kittiwake: This works functionally, but if validation fails the user may receive a misleading message like "Invalid recipe id" for a reviewId field. Consider renaming the function to validatePositiveIntId or introducing a dedicated validateReviewId.

el-kittiwake (changes requested):

greyear:
Fixed the issues mentioned

el-kittiwake (approved):
Thanks, now it works fine!

PR #204, opened 2026-04-13
Auth aware header footer profile navigation

### LINKED ISSUE #142: Set login state to header and footer

SeelaSalorinta:
Changes

- track authentication state in the shared layout
- restore auth state on initial load via /profile
- update header and footer to render based on auth state
- added a basic profile page route and profile fetch
- support links in IconButton

 How auth state is restored

- on page load, the layout checks if the user can access the profile endpoint
- if the request succeeds, the user is treated as signed in
- if it fails, the app stays in the signed-out state

Logout is not implemented yet. For now, the session can be cleared by deleting the auth cookie manually in browser developer tools.

close #142


frontend/app/components/buttons/IconButton.tsx
```diff
@@ -1,24 +1,70 @@
...
+	onClick,
+	onMouseDown,
+	"aria-expanded": ariaExpanded,
+	"aria-haspopup": ariaHaspopup,
+	"aria-label": ariaLabel,
+	"aria-pressed": ariaPressed,
+	"data-initial-focus": dataInitialFocus,
+	disabled = false,
+	type = "button",
 }: IconButtonProps) => {
```
FPyMEHTAPIU: Seems a bit overload now. Let's use something like this:

```suggestion
type BaseProps = {
	children: ReactNode;
	variant?: IconButtonVariants;
	className?: string;
};

type ButtonIconButtonProps = BaseProps &
	Omit<ComponentPropsWithoutRef<"button">, keyof BaseProps> & {
		to?: never;
		"data-initial-focus"?: boolean;
	};

type LinkIconButtonProps = BaseProps &
	Omit<ComponentPropsWithoutRef<typeof Link>, keyof BaseProps | "to" | "aria-pressed"> & {
		to: string;
	};

type IconButtonProps = ButtonIconButtonProps | LinkIconButtonProps;

export const IconButton = (props: IconButtonProps) => {
```

frontend/app/components/buttons/IconButton.tsx
```diff
@@ -1,24 +1,70 @@
...
+			onMouseDown={onMouseDown}
+			aria-expanded={ariaExpanded}
+			aria-haspopup={ariaHaspopup}
+			aria-label={ariaLabel}
+			aria-pressed={ariaPressed}
+			data-initial-focus={dataInitialFocus}
+			disabled={disabled}
 		>
 			{children}
 		</button>
```
FPyMEHTAPIU:
```suggestion
if (props.to !== undefined) {
		const { to, variant: _v, className: _c, children: _ch, ...linkProps } = props;
		return (
			<Link to={to} className={combinedClasses} {...linkProps}>
				{children}
			</Link>
		);
	}

	const {
		to: _to,
		"data-initial-focus": dataInitialFocus,
		type = "button",
		variant: _v,
		className: _c,
		children: _ch,
		...buttonRest
	} = props;
	return (
		<button
			type={type}
			className={combinedClasses}
			data-initial-focus={dataInitialFocus}
			{...buttonRest}
		>
			{children}
		</button>
```

frontend/app/layouts/Footer.tsx
```diff
@@ -34,9 +35,11 @@ export const Footer = ({ onOpenAuthModal }: FooterProps) => {
 					</ul>
 				</nav>
 				<div className="footer-buttons-row">
-					<MainButton onClick={onOpenAuthModal}>
-						{t("common.signInButton")}
-					</MainButton>
+					{isAuthenticated ? null : (
```
FPyMEHTAPIU: Can be simplified a bit:
```suggestion
					{!isAuthenticated && (
```

frontend/app/layouts/Header.tsx
```diff
@@ -48,17 +55,29 @@ const NavigationList = () => {
...
+						<TextIconButton
+							size="body3"
+							to="/profile"
+							variant="inverted"
+							selected={pathname.startsWith("/profile")}
+						>
+							{t("layout.profile")}
+						</TextIconButton>
+					</li>
+				) : null}
```
FPyMEHTAPIU: Same here:

```suggestion
				{isAuthenticated && !isDesktop && (
					<li>
						<TextIconButton
							size="body3"
							to="/profile"
							variant="inverted"
							selected={pathname.startsWith("/profile")}
						>
							{t("layout.profile")}
						</TextIconButton>
					</li>
				)}
```

frontend/app/layouts/Header.tsx
```diff
@@ -125,19 +144,31 @@ export const Header = ({ onOpenAuthModal }: HeaderProps) => {
...
-					<NavigationList />
-					<MainButton variant="inverted" onClick={onOpenAuthModal}>
-						{t("common.signInButton")}
-					</MainButton>
+					<NavigationList isAuthenticated={isAuthenticated} />
+					{isAuthenticated ? null : (
+						<MainButton variant="inverted" onClick={onOpenAuthModal}>
+							{t("common.signInButton")}
+						</MainButton>
+					)}
```
FPyMEHTAPIU: And here:

```suggestion
					{!isAuthenticated && (
						<MainButton variant="inverted" onClick={onOpenAuthModal}>
							{t("common.signInButton")}
						</MainButton>
					)}
```

frontend/app/routes/profile.tsx
```diff
@@ -0,0 +1,94 @@
...
+		fetch(`${API_BASE_URL}/profile`, {
+			credentials: "include",
+		})
+			.then((res) => {
+				if (!res.ok) {
+					setErrorStatus(res.status);
+					return null;
+				}
+
+				return res.json() as Promise<{ data?: ProfileData }>;
```
FPyMEHTAPIU: Let's not have type assertions. Add a proper check before. You can also install Zod.

frontend/app/routes/profile.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+	if (!profile) {
+		return <p className="profile-page-status">{t("profilePage.notFound")}</p>;
+	}
+
+	return (
+		<section className="profile-page">
+			<header className="profile-page-header">
+				<h1>My profile</h1>
+				<p className="text-body2">Quick check for the profile endpoint.</p>
```
FPyMEHTAPIU: These values are not translated

frontend/app/routes/profile.tsx
```diff
@@ -0,0 +1,94 @@
...
+				</p>
+				<p>
+					<span className="text-label">User ID</span>
+					{profile.id}
+				</p>
+				<p>
+					<span className="text-label">Avatar</span>
+					{profile.avatar ?? "No avatar set"}
+				</p>
+			</div>
```
FPyMEHTAPIU, line 118: No translations here as well.

FPyMEHTAPIU (commented):
Hey, great job! 🙂 
This is quite complex thing to handle, so I asked Claude to review it as well and I'll paste it here. I'll also push one commit above with changes suggested by Claude, but I don't really like them all. Some of them are good, some of them are really complicated. Review them and decide what would be simpler and reliable 😉 

Claude review:
`frontend/app/routes/profile.tsx`
1. as type assertion (violates project guidelines)

`return res.json() as Promise<{ data?: ProfileData }>;`

`res.json()` returns `Promise<unknown>`. Casting with as bypasses type safety. Use a type guard instead:
```
function isProfileBody(v: unknown): v is { data?: ProfileData } {
  return typeof v === "object" && v !== null && "data" in v;
}
// then:
const body: unknown = await res.json();
if (!isProfileBody(body)) { ... }
```
2. Missing `useEffect` cleanup flag

If the component unmounts mid-fetch, the `.then()` callbacks still run and call `setProfile` / `setIsLoading` on an unmounted component. Add an ignore flag:
```
useEffect(() => {
  let ignore = false;
  fetch(...)
    .then(...) // only call setState if !ignore
  return () => { ignore = true; };
}, []);
```
3. Hardcoded English strings

The profile content uses hardcoded strings instead of `t()`:

"My profile", "Quick check for the profile endpoint.", "Username", "User ID", "Avatar", "No avatar set"
These should either use the i18n system or the page should be marked clearly as a placeholder. The comment "Quick check for the profile endpoint" suggests this is intentionally temporary — if so, a TODO comment would make intent explicit.

`frontend/app/layouts/layout.tsx`
4. Auth state fetched twice

`layout.tsx` fetches `/profile` to check authentication, then `profile.tsx` independently fetches `/profile` again when the user navigates there. A shared auth context (or passing the already-fetched data down) would avoid the duplicate request.

5. No sign-out path

`isAuthenticated` can only go false → true (on login success). There's no mechanism to go true → false (on logout). When a sign-out feature is added to the profile page, it would need a way to update this state, which the current prop-drilling pattern doesn't support. An auth context would handle this more cleanly.

`frontend/app/components/buttons/IconButton.tsx`
6. Explicit ARIA prop list is brittle

Replacing ...props with an explicit allowlist of ARIA attributes means every new attribute (e.g., aria-controls, aria-describedby) requires updating the type definition, destructuring, and JSX spread in both branches. A discriminated union would keep the flexibility:
```
type ButtonProps = { to?: never; disabled?: boolean; type?: ...; } & ButtonHTMLAttributes<HTMLButtonElement>;
type LinkProps   = { to: string } & LinkProps_from_react_router;
type IconButtonProps = (ButtonProps | LinkProps) & { children: ReactNode; variant?: ...; className?: string; };
```
7. aria-pressed passed to `<Link>`

`aria-pressed` is a button-specific state attribute. Passing it to a `<Link>` (which renders as `<a>`) is semantically incorrect. The Link branch should omit it.

Translations
8. `signedInButton` key added but never used

All three locale files add "signedInButton" but it is not referenced anywhere in the codebase. Either wire it up or remove it to avoid dead translation keys.

SeelaSalorinta:
i tried to implement pretty much all the suggestions, but left out the auth/context refactor since it felt like a bigger structural change than this pr needs imo

FPyMEHTAPIU (approved):

PR #205, opened 2026-04-13
Adding status change logic (BE part)

greyear:
- Adding new field to the users db (last_seen_at)
- Adding middleware to update last_seen_at for every incoming request
- Adding routes + services for it

 A user is considered online if their last activity was within **60 seconds**.
Status update is happening in getUserById handler for GET request

Endpoint for FE
POST /users/me/heartbeat

### DIFF
--- a/backend/services/core-service/src/app.ts
+++ b/backend/services/core-service/src/app.ts
@@ -12,6 +12,8 @@ import path from "node:path";
 import cors from "cors";
 import express, { type Express } from "express";
 import { errorHandler, notFoundHandler } from "./middleware/errorHandler.js";
+import { extractUser } from "./middleware/extractUser.js";
+import { updateLastSeen } from "./middleware/updateLastSeen.js";
 import { healthRouter } from "./routes/health.routes.js";
 import { profileRouter } from "./routes/profile.routes.js";
 import { recipesRouter } from "./routes/recipes.routes.js";
@@ -44,8 +46,12 @@ app.use(
 app.use(express.json());
 // Serve uploaded avatars as static files at /avatars/*
 app.use("/avatars", express.static(path.resolve("uploads/avatars")));
-// Serve uploaded pictures as static files at /recipes/*
+// Serve uploaded pictures as static files at /recipe-pictures/*
 app.use("/recipe-pictures", express.static(path.resolve("uploads/recipes")));
+// Extract userId from X-User-Id header on every request
+app.use(extractUser);
+// Update last_seen_at on every authenticated request
+app.use(updateLastSeen);
 
 // ===== ROUTES =====
 // /health/* → healthRouter
--- a/backend/services/core-service/src/routes/users.routes.ts
+++ b/backend/services/core-service/src/routes/users.routes.ts
@@ -352,7 +352,6 @@ const heartbeatHandler = async (
             error.statusCode = 401;
             throw error;
         }
-        await updateHeartbeat(req.userId);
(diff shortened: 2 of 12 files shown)


backend/services/core-service/src/routes/users.routes.ts
```diff
@@ -340,7 +341,26 @@ const getFollowingHandler = async (
...
+    res: Response,
+    next: NextFunction,
+): Promise<void> => {
+    try {
+        if (req.userId === undefined) {
+            const error: CustomError = new Error("Authentication required");
+            error.statusCode = 401;
+            throw error;
+        }
+        await updateHeartbeat(req.userId);
```
FPyMEHTAPIU: Do we need it? Seems like `updateLastSeen` middleware handles it.

backend/services/core-service/src/services/users.service.ts
```diff
@@ -170,11 +56,26 @@ const parseUserRows = <T>(
...
+ * Update last_seen_at for a user.
+ * Called explicitly by POST /users/me/heartbeat.
+ * Also called implicitly by updateLastSeen middleware on every authenticated request.
+ */
+export const updateHeartbeat = async (userId: number): Promise<void> => {
+	await pool.query(
+		`UPDATE users SET last_seen_at = now() WHERE id = $1`,
+		[userId],
+	);
+};
```
FPyMEHTAPIU: Not sure if we even need it. As I mentioned previously, `updateLastSeen` probably handles this logic.

backend/services/core-service/src/app.ts
```diff
@@ -44,8 +46,12 @@ app.use(
 app.use(express.json());
 // Serve uploaded avatars as static files at /avatars/*
 app.use("/avatars", express.static(path.resolve("uploads/avatars")));
-// Serve uploaded pictures as static files at /recipes/*
+// Serve uploaded pictures as static files at /recipe-pictures/*
 app.use("/recipe-pictures", express.static(path.resolve("uploads/recipes")));
+// Extract userId from X-User-Id header on every request
+app.use(extractUser);
```
FPyMEHTAPIU, line 52: Since you applied `extractUser` to the app, you might want to remove explicit middlewares from the endpoints in other files.

FPyMEHTAPIU (commented):
Nice work! 💪 
Seems like there're some redundant calls, let's verify them.

backend/services/core-service/src/routes/users.routes.ts
greyear: Yes, we need it in case user just stays on the website idle. We need to update last seen so we won't have a scenario when user stays on the page inactive and after 1 min becomes offline.

backend/services/core-service/src/services/users.service.ts
greyear: Pls check comment above

backend/services/core-service/src/routes/users.routes.ts
FPyMEHTAPIU: but in `app.ts` there's 
```
app.use(updateLastSeen);
```
which changes the lastSeen during every request.

greyear: I could remove it from services and just make it return 200 since we need it anyway to trigger middleware

greyear:
@FPyMEHTAPIU could you pls review again, adjusted some stuff with discussed

FPyMEHTAPIU (approved):

PR #209, opened 2026-04-14
http->https BE gateway migration

greyear:
Adding traefik as a reverse proxy. It:

- Accepting HTTPS on port 443 using your self-signed cert
- Redirecting HTTP (port 80) to HTTPS permanently
- Forwarding requests to api-gateway internally

New rule to Makefile added to create self signed cert. New container for Traefik added.

!!! Important info for FE: starting from now FE should send requests to https://localhost/ instead of http://localhost:3000/

I've changed that part:

export const API_BASE_URL =
	import.meta.env.VITE_API_BASE_URL ?? "https://localhost";

But just in case


scripts/generate-certs.sh
```diff
@@ -0,0 +1,20 @@
...
+echo "Generating self-signed certificates..."
+
+mkdir -p certs
+
+openssl req -x509 -newkey rsa:4096 \
+    -keyout certs/key.pem \
+    -out certs/cert.pem \
+    -days 365 -nodes \
+    -subj "/CN=localhost"
+
```
FPyMEHTAPIU: Claude:

> Modern TLS clients (browsers, Node.js, curl without -k) ignore the CN field and only validate against Subject Alternative Names. Without a SAN, any client doing proper certificate verification rejects the cert with an error like "certificate has no SANs matching the host name". The cert was being generated but would silently fail verification everywhere except scripts using -k.
> 
> Note: if you already have a certs/ directory from a previous run, delete it and re-run make certs — the script skips generation if the files already exist, so the old broken cert won't be replaced otherwise.


```suggestion
    -subj "/CN=localhost" \
    -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
```

docker-compose.yml
```diff
@@ -74,6 +92,18 @@ services:
...
+      - "traefik.enable=true"
+      - "traefik.http.routers.gateway-http.rule=Host(`localhost`)"
+      - "traefik.http.routers.gateway-http.entrypoints=web"
+      - "traefik.http.routers.gateway-http.middlewares=redirect-to-https"
+      - "traefik.http.routers.gateway-https.rule=Host(`localhost`)"
+      - "traefik.http.routers.gateway-https.entrypoints=websecure"
+      - "traefik.http.routers.gateway-https.tls=true"
+      - "traefik.http.services.gateway.loadbalancer.server.port=3000"
+      - "traefik.http.middlewares.redirect-to-https.redirectscheme.scheme=https"
+      - "traefik.http.middlewares.redirect-to-https.redirectscheme.permanent=true"
```
FPyMEHTAPIU: Claude:

> traefik.yml already handles the HTTP→HTTPS redirect at the entrypoint level — all traffic on port 80 is redirected globally before it reaches any router. The router-level middleware was dead code on top of that. Having both also means a gateway-http router was registered that referenced a middleware defined on the same router, which is redundant at best and confusing at worst.


```suggestion
      - "traefik.enable=true"
      - "traefik.http.routers.gateway-https.rule=Host(`localhost`)"
      - "traefik.http.routers.gateway-https.entrypoints=websecure"
      - "traefik.http.routers.gateway-https.tls=true"
      - "traefik.http.services.gateway.loadbalancer.server.port=3000"
```

test-endpoints.sh
```diff
@@ -34,7 +34,7 @@ check_endpoint() {
   local url="$BASE_URL$path"
   local actual_status
   
-  actual_status=$(curl -sS -o /tmp/test-resp.json -w "%{http_code}" "$url" 2>/dev/null || echo "000")
+  actual_status=$(curl -sS -k -o /tmp/test-resp.json -w "%{http_code}" "$url" 2>/dev/null || echo "000")
```
FPyMEHTAPIU: Claude:

> -k disables all TLS verification, meaning the tests would pass even if the cert was wrong, expired, or from the wrong host. --cacert certs/cert.pem tells curl to trust exactly the self-signed cert we generated — it verifies the chain and hostname (now that the SAN is present) without requiring a system-trusted CA. The auth test scripts had no TLS flag at all and would have thrown SSL errors after the https:// URL switch; those were also updated.
> 
> All scripts assume they're run from the project root (consistent with how the Makefile invokes them), which is where certs/ lives.


```suggestion
  actual_status=$(curl -sS --cacert certs/cert.pem -o /tmp/test-resp.json -w "%{http_code}" "$url" 2>/dev/null || echo "000")
```

Same in other places.

FPyMEHTAPIU (dismissed):
Looks nice to me. 🫶 
Left some comments generated by Claude, you decide fix them or not. Don't really know much about it 😓

scripts/generate-certs.sh
```diff
@@ -16,5 +16,6 @@ openssl req -x509 -newkey rsa:4096 \
     -out certs/cert.pem \
     -days 365 -nodes \
     -subj "/CN=localhost"
```
FPyMEHTAPIU: `\` is missing here, won't work

FPyMEHTAPIU (changes requested):
Everything is nice, but `\` is missing at the end of the line, won't work without it

FPyMEHTAPIU (approved):

PR #210, opened 2026-04-15
Auth week16 changes agreed meetings

el-kittiwake:
Changed /delete and /change-password to require userId as request body.
Change /register and /login to not require username. Only email and password.
Changed the logic of /google a bit to remove what seems like an unnecessary call to itself.

Added /auth/me endpoint for @SeelaSalorinta. Takes token, returns userId and email.
Added PUT to /profile for core-service. Notifies that a new user has been created.

A lot of Biome alterations all over the backend, hopefully nothing harming anything. If I had known it would change more than mine I would have left it alone.

Updated documentation for auth-endpoints and auth-tests.

Manual and AI generated tests seem to return what they should.


el-kittiwake:
Fixed merge conflicts.

el-kittiwake:
This should also resolve #220

greyear (approved):
Thank you! Looks like it works properly!

PR #217, opened 2026-04-15
Users/id/favorites

### LINKED ISSUE #215: Implement GET /users/:id/favorites
Returns list of published recipes favorited by the specified user only if mutual follow relationship exists.

greyear:
Implements GET /users/:id/favorites
- Requires authentication (X-User-Id from API Gateway).
- Access denied if users are not mutual followers (returns 403).
- Returns list of published recipes favorited by the specified user only if mutual follow relationship exists.
Closes #215


backend/services/core-service/src/services/recipes.service.ts
```diff
@@ -875,6 +879,72 @@ export const getMyFavoriteRecipes = async (
...
+		);
+
+		if (userCheckResult.rows.length === 0) {
+			const error: any = new Error("User not found");
+			error.code = "USER_NOT_FOUND";
+			throw error;
+		}
+
+		// Check if current user and target user are mutual followers
+		const mutualFollowResult = await pool.query(
```
el-kittiwake: Let's reuse with areMutualFollowers() func

el-kittiwake (approved):
Looks good now, thanks

PR #222, opened 2026-04-17
Recipe creating page

### LINKED ISSUE #64: Create recipe creating page

FPyMEHTAPIU:
This is so big PR, sorry for that. 😿 
I decided not to split all the functionality to a separate PR, which would take more time and effort for refactoring.

New `/recipes/create` route with a full recipe authoring form: photo upload (with preview), title, description (with live character counter), servings, prep/cook times, drag-and-drop ingredient rows, and drag-and-drop instruction steps. Wired the "Create" button on the recipes page to navigate here.

### Highlights

- Drag-and-drop via `@hello-pangea/dnd` for ingredients and instructions
- Zod v4 schema validation on submit (numbers coerced via `z.preprocess`, cross-field refinements for prep/cook totals)
- Shared accessible `TextArea` component with blur-based required validation, `aria-invalid`/`aria-describedby`, and an optional `footer` slot (used for the description counter)
- Semantic `<fieldset>`/`<legend>` grouping for ingredient rows and time inputs
- Deterministic initial row ids via `useId()` to avoid SSR hydration mismatches; `crypto.randomUUID()` only for dynamically added rows
- Ingredient row DOM order matches visual order across breakpoints (drag → name → amount → unit → remove) so keyboard tab navigation flows left-to-right, top-to-bottom on both mobile and tablet/desktop; fields top-aligned so error messages don't push siblings out of line

Submit handler is wired and validated; API call is left as a TODO.

closes #64


SeelaSalorinta (approved):
Works and looks nice in browser! Excellent Nick! 💯 

 i also asked AI to review this bc so long pr, so i approved and you can fix anything if you feel like its needed

Findings

Medium: Required text fields accept whitespace-only values.
The schema currently uses .min(1) for fields like title, description, ingredient name, and instruction text, so values like " " still pass validation. Please trim before validating, for example with .trim().min(1, ...), or preprocess the input before schema validation.

Medium: Ingredient amount validation is inconsistent between the native number input and the Zod schema.
The input uses type="number" with min={0} and no step, while the schema requires .positive() and appears to allow decimals. This creates a mismatch where 0 can look acceptable until submit, while something like 0.5 may trigger browser step validation even though the schema accepts it. Please align the input and schema behavior, e.g. min={0.01} with step="any", or make the schema integer-only if decimals are not intended.

Medium: Drag handles do not have an accessible name.
The reorder handles appear to be focusable through drag handle props, but the visible icon is decorative/hidden from assistive tech, which leaves the control unnamed for screen reader users. Please add explicit accessible labels such as aria-label="Reorder ingredient 1" / aria-label="Reorder step 1".

Low/Medium: The new create flow bypasses the existing i18n pattern.
This route and several new child components hard-code visible labels, placeholders, validation copy, and ARIA labels instead of using the existing translation setup. That means locale switching will leave this whole flow in English. Please move new user-facing strings into translations for consistency with the rest of the app.

Open question

Should recipe creation require authentication?
The Create button now links directly to /recipes/create, but from this PR it’s not clear that the route is gated the same way as other user actions such as rating. If anonymous recipe creation is not intended, this should probably be protected before merge.

FPyMEHTAPIU:
Hey! 
There're some good points (accessibility and auth for creating, but I won't put them there now, since it's already huge and I'll add more connectivity for it in the following PR. Thank you 😍

PR #230, opened 2026-04-17
Pagination be

greyear:
Implemented:
GET /recipes?lang=en&page={}&per_page={}

Body sent back to FE:
{
	data,
	total_pages,
	total_count,
	page,
	per_page
}

Schema:
PaginatedResponse<T> = 
{
	data: T[];
	total_count: number;
	total_pages: number;
	page: number;
	per_page: number;
};

### DIFF
--- a/backend/services/core-service/src/routes/recipes.routes.ts
+++ b/backend/services/core-service/src/routes/recipes.routes.ts
@@ -26,7 +26,6 @@ import {
 	archiveRecipe,
 	createRecipe,
 	deleteReview,
-	getAllRecipes,
 	getAllRecipesPaginated,
 	getRecipeById,
 	getRecipeReviews,
@@ -48,6 +47,7 @@ import {
 	validateReviewId,
 	validateUpdateRecipeInput,
 	validateUpdateRecipeReviewInput,
+	validatePaginationQuery,
 } from "../validation/schemas.js";
 import { ratingsRouter } from "./ratings.routes.js";
 
@@ -489,13 +489,12 @@ const getAllRecipesHandler = async (
 	try {
 		const locale = resolveRequestedLocale(req);
 
-		const page = Math.max(1, Number.parseInt(req.query.page as string) || 1);
-		const perPage = Math.min(
-			100,
-			Math.max(1, Number.parseInt(req.query.per_page as string) || 12),
-		);
-
-		const result = await getAllRecipesPaginated(page, perPage, locale);
+		const pagination = validatePaginationQuery(req.query);
+		if (!pagination.valid) {
+			res.status(400).json({ error: pagination.error });
+			return;
+		}
+		const result = await getAllRecipesPaginated(pagination.value.page, pagination.value.per_page, locale);
 		res.status(200).json(result);
 	} catch (error) {
 		next(error);
--- a/backend/services/core-service/src/routes/recipes.routes.ts
+++ b/backend/services/core-service/src/routes/recipes.routes.ts
@@ -27,6 +27,7 @@ import {
 	createRecipe,
 	deleteReview,
 	getAllRecipes,
+	getAllRecipesPaginated,
 	getRecipeById,
 	getRecipeReviews,
 	leaveRecipeReview,
@@ -476,7 +477,7 @@ const unfavoriteRecipeHandler = async (
 };
 
 /**
- * GET /recipes - fetch all published recipes
+ * GET /recipes - fetch all published recipes with pagination
  *
  * Note: No authentication needed - returns only published recipes
  */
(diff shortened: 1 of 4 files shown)


backend/services/core-service/src/routes/recipes.routes.ts
```diff
@@ -27,6 +27,7 @@ import {
 	createRecipe,
 	deleteReview,
 	getAllRecipes,
+	getAllRecipesPaginated,
```
FPyMEHTAPIU:
```suggestion
	getAllRecipesPaginated,
```

FPyMEHTAPIU: getAllRecipesPaginated should replace the old getter, right?

backend/services/core-service/src/routes/recipes.routes.ts
```diff
@@ -487,8 +488,15 @@ const getAllRecipesHandler = async (
...
-		const recipes = await getAllRecipes(locale);
-		res.status(200).json({ data: recipes, count: recipes.length });
+
+		const page = Math.max(1, Number.parseInt(req.query.page as string) || 1);
+		const perPage = Math.min(
+			100,
+			Math.max(1, Number.parseInt(req.query.per_page as string) || 12),
+		);
+
+		const result = await getAllRecipesPaginated(page, perPage, locale);
```
FPyMEHTAPIU: This is very fragile and dangerous (I mean the "as" type assertion). Let's have proper Zod validation instead.
```suggestion
		const pagination = validatePaginationQuery(req.query);
		if (!pagination.valid) {
			res.status(400).json({ error: pagination.error });
			return;
		}

		const result = await getAllRecipesPaginated(pagination.value.page, pagination.value.per_page, locale);
```

FPyMEHTAPIU: Zod schema is in `schemas.ts`.

backend/services/core-service/src/services/recipes.service.ts
```diff
@@ -420,6 +420,57 @@ export const getAllRecipes = async (
 	}
 };
 
+export type PaginatedRecipes = {
+	data: RecipeListItem[];
+	total_count: number;
+	total_pages: number;
+	page: number;
+	per_page: number;
+};
```
FPyMEHTAPIU: You can remove these lines (yes, including the type since it's validated by Zod). Paginated recipes does basically the same.

backend/services/core-service/src/services/recipes.service.ts
```diff
@@ -420,6 +420,57 @@ export const getAllRecipes = async (
...
+
+/**
+ * Get published recipes for exact page
+ */
+export const getAllRecipesPaginated = async (
+	page: number,
+	perPage: number,
+	locale: SupportedLocale = DEFAULT_LOCALE,
+): Promise<PaginatedRecipes> => {
```
FPyMEHTAPIU:
```suggestion
): Promise<PaginatedResponse<RecipeListItem>> => {
```

backend/services/core-service/src/validation/schemas.ts
```diff
@@ -412,6 +412,23 @@ export type FavoriteRecipeListItem = z.infer<
...
 >;
 
+export const PaginatedResponseSchema = <T extends z.ZodTypeAny>(itemSchema: T) =>
+	z.object({
+		data: z.array(itemSchema),
+		total_pages: z.number().int().nonnegative(),
+		total_count: z.number().int().nonnegative(),
+		page: z.number().int().positive(),
+		per_page: z.number().int().positive(),
+	});
```
FPyMEHTAPIU: it seems unused.
```suggestion
```

backend/services/core-service/src/validation/schemas.ts
```diff
@@ -412,6 +412,23 @@ export type FavoriteRecipeListItem = z.infer<
...
+
+export type PaginatedResponse<T> = {
+	data: T[];
+	total_count: number;
+	total_pages: number;
+	page: number;
+	per_page: number;
+};
+
```
FPyMEHTAPIU, line 422: you can extend the validation in a following way:

```suggestion

const paginationQuerySchema = z.object({
	page: z.coerce.number().int().positive().default(1),
	per_page: z.coerce.number().int().min(1).max(100).default(12),
});

export type PaginationQuery = z.infer<typeof paginationQuerySchema>;

export const validatePaginationQuery = (
	input: unknown,
): ValidationResult<PaginationQuery> => {
	const result = paginationQuerySchema.safeParse(input);

	if (result.success) {
		return { valid: true, value: result.data };
	}

	return { valid: false, error: result.error.issues[0]?.message ?? "Invalid pagination parameters" };
};
```

FPyMEHTAPIU (changes requested):
Hey, good job!
It has validation issue and some duplicated code. Otherwise, looks good!

FPyMEHTAPIU (dismissed):

FPyMEHTAPIU (approved):

PR #237, opened 2026-04-17
add frontend Google login flow

### LINKED ISSUE #195: Implement Google login flow

SeelaSalorinta:
Whats included:
- loads Google Identity Services script in AuthForm
- initializes Google with VITE_GOOGLE_CLIENT_ID
- renders the official Google button (switches between sign in / sign up based on mode)
- sends the Google credential to /auth/google
- adds handling for loading, missing config, and error states
- adds VITE_GOOGLE_CLIENT_ID to frontend env example
- rename loginPage translation namespace to authModal

Prerequisites
- Frontend and backend env google client variables set

close #195

### DIFF
--- a/frontend/app/assets/styles/auth.css
+++ b/frontend/app/assets/styles/auth.css
@@ -21,21 +21,41 @@
 	background-color: var(--surface-card);
 	border: 1px solid var(--borders-action);
 	color: var(--text-primary);
-}
-
-.auth-social-button-content {
-	display: flex;
-	align-items: center;
-	justify-content: center;
-	gap: var(--space-s);
 	width: 100%;
+	pointer-events: none;
 }
 
 .auth-social-google {
-	display: inline-flex;
+	stroke-width: 1;
+}
+
+.auth-google-button-shell {
+	position: relative;
+	display: flex;
+	justify-content: center;
+	width: 100%;
+}
+
+.auth-google-render-target {
+	position: absolute;
+	inset: 0;
+	z-index: 2;
+	display: flex;
 	align-items: center;
 	justify-content: center;
-	stroke-width: 1;
+	width: 100%;
+	opacity: 0;
+}
+
+.auth-google-render-target[aria-hidden="true"] {
+	visibility: hidden;
+	pointer-events: none;
+}
+
+.auth-google-render-target > *,
+.auth-google-render-target iframe {
+	width: 100%;
+	height: 100%;
 }
 
 .auth-form {
@@ -73,7 +93,6 @@
 	display: flex;
 	flex-direction: column;
 	align-items: flex-start;
-	gap: var(--space-xs);
 }
 
 .auth-modal-close-row {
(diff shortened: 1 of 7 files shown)


frontend/app/assets/styles/auth.css
FPyMEHTAPIU, line 1: in mobile 328px it overflows. Let's have line wrap.

Image

frontend/app/components/auth/AuthForm.tsx
```diff
@@ -10,28 +18,201 @@ import "~/assets/styles/auth.css";
...
+		if (window.google?.accounts?.id) {
+			resolve();
+			return;
+		}
+
+		const existingScript = document.querySelector<HTMLScriptElement>(
+			`script[src="${GOOGLE_IDENTITY_SCRIPT_SRC}"]`,
+		);
+		if (existingScript) {
+			existingScript.addEventListener("load", () => resolve(), { once: true });
```
FPyMEHTAPIU, line 85: A data-gsi-status attribute ("loading" --> "loaded" / "error") is now written to the <script> tag as it progresses. When a subsequent call finds an existing tag, it reads this status and either resolves or rejects synchronously — instead of attaching event listeners that will never fire because the events already happened. The "loading" case (script still in flight) falls through to the existing listener logic unchanged.
```suggestion
		if (existingScript) {
		const status = existingScript.dataset.gsiStatus;
			if (status === "loaded") {
				resolve();
				return;
			}
			if (status === "error") {
				reject(new Error("Google Identity Services failed to load"));
				return;
			}
			existingScript.addEventListener("load", () => resolve(), { once: true });
```

frontend/app/components/auth/AuthForm.tsx
```diff
@@ -10,28 +18,201 @@ import "~/assets/styles/auth.css";
...
+			return;
+		}
+
+		const script = document.createElement("script");
+		script.src = GOOGLE_IDENTITY_SCRIPT_SRC;
+		script.async = true;
+		script.defer = true;
+		script.onload = () => resolve();
+		script.onerror = () =>
+			reject(new Error("Google Identity Services failed to load"));
```
FPyMEHTAPIU:
```suggestion
		script.dataset.gsiStatus = "loading";
		script.onload = () => {
			script.dataset.gsiStatus = "loaded";
			resolve();
		};
		script.onerror = () => {
			script.dataset.gsiStatus = "error";
			reject(new Error("Google Identity Services failed to load"));
		};
```

frontend/app/components/auth/AuthForm.tsx
```diff
@@ -175,35 +361,39 @@ export const AuthForm = ({
...
-						variant="inverted"
-						className="auth-social-button"
-						onClick={handleGoogleLogin}
-					>
-						<span className="auth-social-button-content">
+					<div className="auth-google-button-shell">
+						<button
+							type="button"
+							className="main-button inverted auth-social-button"
+							disabled={!canUseGoogleButton}
```
FPyMEHTAPIU: This seems like a "fake" button, so it can probably be disabled unconditionally.  
```suggestion
							disabled
```

FPyMEHTAPIU (commented):
Didn't have much time for testing, let's do it today at school.
Left some comments and would like to see the overflow fix.
Otherwise, seems nice 💪

FPyMEHTAPIU (approved):

PR #238, opened 2026-04-18
Attempt to fix "Wait for API Gateway to be ready" smoke test failure

el-kittiwake:
Change to api-gateway section of docker-compose.yml.

ports rather than expose because curl.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -66,8 +66,6 @@ services:
     ports:
       - "127.0.0.1:3000:3000"
 
-    #expose:
-    #  - "3000"
 
     environment:
       NODE_ENV: development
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -63,8 +63,11 @@ services:
     container_name: api-gateway
     restart: unless-stopped
 
-    expose:
-      - "3000"
+    ports:
+      - "127.0.0.1:3000:3000"
+
+    #expose:
+    #  - "3000"
```
FPyMEHTAPIU: This should be removed instead.
```suggestion
```

FPyMEHTAPIU (dismissed):
Looks nice, but I'd like to remove the commented part.

greyear (changes requested):
Didn't work for me on GH Actions page

el-kittiwake:
There still seems to be failures after the one this is attempting to fix.

greyear:
We'll try to solve it diferently

PR #250, opened 2026-04-18
Connect favorite button to be

### LINKED ISSUE #241: Connect recipe favorite heart to backend

SeelaSalorinta:
Summary
- Add favorite functionality for recipes across the app
- Load the current user’s favorite recipes into recipe grids
- Connect favorite buttons to backend favorite/unfavorite endpoints
- Update the heart immediately when clicked, and undo it only if the backend request fails
- Keep the selected heart state when the backend says the recipe is already favorited or already unfavorited
- Disable favorite buttons while favorites are loading
- Opens authModal when heart is clicked if user is not signed in, after successful login adds automatically to favorites

Note
- Profile page currently just renders favorites for testing
- Commit history is a bit messy (accidentally pushed via Jimi’s account and had issues squashing), so left as-is

close #241


frontend/app/components/RecipesGrid.tsx
```diff
@@ -19,8 +20,18 @@ type RecipesGridProps = {
 	perPage?: number;
 	onLoad?: (totalCount: number) => void;
 	sort?: "top";
+	isAuthenticated: boolean;
+	openAuthModal: (onSuccessAction?: () => void) => void;
 };
 
+const FavoriteRecipeSchema = z.object({
```
greyear: This ```FavoriteRecipeSchema``` schema been initialised and used in multiple files. Should we move it to the separate file?

frontend/app/routes/profile.tsx
```diff
@@ -80,7 +104,58 @@ const ProfilePage = () => {
 		};
 	}, []);
 
-	if (isLoading) {
+	useEffect(() => {
```
greyear, line 102: Should we put a protection from 401?
Like you have in other files:
```if (!isAuthenticated)```

File: frontend/app/routes/home.tsx [hunk 3]
```diff
@@ -51,7 +55,14 @@ const HomePage = () => {
 						<NavArrowRight aria-hidden="true" />
 					</TextIconButton>
 				</div>
-				<RecipesGrid sort="top" page={1} perPage={recipesPerPage} />
+				<RecipesGrid
+					sort="top"
+					sortValue=""
```
greyear: From AI review. Pls check the necessity of this change.

RecipesGridProps requires sortValue, but the home page passes an empty string because it has no sort UI — it uses sort="top" instead. Either make sortValue optional (default "") or document that sort="top" ignores sortValue. As-is it's a leaky abstraction.

frontend/app/components/RecipesGrid.tsx
```diff
@@ -85,6 +130,101 @@ export const RecipesGrid = ({
...
+		const shouldBeFavorited = !wasFavoritedBeforeClick;
+
+		setFavoriteIds((prev) =>
+			updateFavoriteIds(prev, recipeId, shouldBeFavorited),
+		);
+
+		setPendingFavoriteIds((prev) => new Set(prev).add(recipeId));
+
+		try {
+			const response = await fetch(
```
greyear, line 170: For this part the optimistic update runs once, and then shouldBeFavorited is re-applied two more times (409 and success). Suggested fix:

```setFavoriteIds((prev) => updateFavoriteIds(prev, recipeId, shouldBeFavorited));
setPendingFavoriteIds((prev) => new Set(prev).add(recipeId));

try {
    const response = await fetch(...);

    if (!response.ok && response.status !== 409) {
        setFavoriteIds((prev) =>
            updateFavoriteIds(prev, recipeId, wasFavoritedBeforeClick),
        );
    }
    // success and 409 both keep the optimistic state — no setState needed
} catch (error) {
    console.error(error);
    setFavoriteIds((prev) =>
        updateFavoriteIds(prev, recipeId, wasFavoritedBeforeClick),
    );
} finally {
    setPendingFavoriteIds((prev) => {
        const next = new Set(prev);
        next.delete(recipeId);
        return next;
    });
}```

greyear (changes requested):
Looks nice!
Suggested some changes, refactoring mostly

greyear (approved):
Thanks! Now looks good!

PR #255, opened 2026-04-19
get rating added to existing endpoint that returns reviews

greyear:
It's the endpoint fix for the FE to fetch data smoothly


el-kittiwake (approved):
Looks good to me. Tested with Postman.

PR #259, opened 2026-04-19
now 20 recipes everywhere and authors have different amount of them

greyear:
1) 20 recipes now instead of 41 (20 were doubled + 1 in the other file)
2) authors have different amounts
3) ingredients added


greyear:
now there are 20 recipes (not 41) with null rating, different authors and JSONB translations

el-kittiwake:
Everything works. Claude's comments could be more proper, but would changing them make anything work better?
From what I can tell, this is fine.

el-kittiwake (approved):
IT all works. Behaves as I expect. I think it is is fine.

PR #260, opened 2026-04-19
Search page

### LINKED ISSUE #125: Create search page

FPyMEHTAPIU:
### Highlights

- New `/search` page with query, recipes/users tabs, sort menu, and filter button scaffold.
- Wired to real backend (GET /search/recipes) — response shape (`recipe_id`, `count`, `summary`) mapped into the existing recipe card model.
- URL-driven state: q, type, sort, limit, ai — reload-safe and shareable.
- Per-page selector (reuses `SortMenu`), placed next to pagination. Capped at backend's `SEARCH_MAX_RESULT_LIMIT`.
- AI summary toggle next to the search field — Sparks icon, aria-pressed, state-aware labels; hides the AI-generated summary when off.
- Loading state (Searching recipes...) and empty-state fallback when no matches.
- i18n for all new strings (en/fi/ru), including aria-labels.
- Minor UX fixes: `SearchField` no longer expands inside the header, unique IDs per field, notification icon removed from header.

closes #125


frontend/app/routes/search.tsx
```diff
@@ -0,0 +1,287 @@
...
+					summary: body.summary,
+					data: body.data.map((item) => ({
+						id: item.recipe_id,
+						title: item.title,
+						description: item.description ?? "",
+						rating_avg: "",
+					})),
+				});
+			})
+			.catch(console.error)
```
SeelaSalorinta: Previous results remain visible if a new search fails
 If a search request fails, the catch only logs the error and does not update results. Because of that, the page keeps showing the previous successful results even though the query has changed.
Example:
- Search “pasta” → success → pasta results shown
- Search “soup” → request fails → loading stops

UI still shows pasta results while the query is “soup”
This can be misleading. It would be better to either clear the results in the catch (e.g. setResults(null)) or add an error state so the UI reflects that the search failed.

frontend/app/routes/search.tsx
```diff
@@ -0,0 +1,287 @@
...
+			limit: String(limit),
+		});
+
+		setIsLoading(true);
+		fetch(`${API_BASE_URL}/search/recipes?${params}`)
+			.then(async (res) => {
+				if (!res.ok) {
+					return { count: 0, data: [] } as SearchRecipesApiResponse;
+				}
+				return (await res.json()) as SearchRecipesApiResponse;
```
SeelaSalorinta: Using as here skips type checking, so mismatches in the API response wouldn’t be caught. Might be worth avoiding assertions or validating the response shape instead.

SeelaSalorinta (approved):
Nice work on this Nick, the structure is really clear and the URL-driven state makes the page feel solid and shareable 🥇  Everything looks really good visually, and from testing the search works well!

Just left a couple small comments:
- about handling failed search requests (to avoid showing stale results)
- and a note on the type assertions in the fetch

Also noticed that rating_avg isn’t shown on recipes in search page ,but i guess that’s related to the backend response.

Overall looks great to me !!🙌

SeelaSalorinta (approved):
Approved!

PR #262, opened 2026-04-20
Env consolidation

greyear:
- Moved all .envs to the global .env file
- Updated templates and docker compose
- Updated jest tests rule in Makefile


SeelaSalorinta (commented):
Just wondering: since PORT is repeated three times here, should these maybe be service-specific keys like API_GATEWAY_PORT, CORE_SERVICE_PORT, and AUTH_SERVICE_PORT instead?
Also, are JWTSecret / JWT_SECRET and NODEENV / NODE_ENV intentionally separate, or are those meant to be the same key?

.env.template
```diff
@@ -1,52 +1,37 @@
...
-SEARCH_REINDEX_TIMEOUT_MS="5000"
+POSTGRES_PORT="1111"
+POSTGRES_DB="db"
+POSTGRES_USER="user"
+POSTGRES_PASSWORD="password"
+
+# auth-service
+MONGODB_AUTH_URI=mongodb://localhost:11111/db
+JWTSecret="jwt_secret_key"
+JWT_SECRET="jwt_secret_key"
```
FPyMEHTAPIU: Duplication? I think we need to keep only `JWT_SECRET`.

.env.template
```diff
@@ -1,52 +1,37 @@
...
+PORT=1111
```
FPyMEHTAPIU: The port is duplicated.

.env.template
```diff
@@ -1,52 +1,37 @@
...
-GOOGLE_CLIENT_ID="123456789012-abc123def456gh789ijklmn0pqrstuvw.apps.googleusercontent.com"
-CORE_SERVICE_URL="http://core-service:3002"
+NODE_ENV="development"
+CORS_ORIGIN="https://localhost:1111"
 
-# Core service
-CORE_SERVICE_PORT="3333"
-DATABASE_URL_CORE="postgresql://your_user:your_password@localhost:5555/your_database"
+# core-service
+NODEENV="development"
```
FPyMEHTAPIU: typo? I assume it should be `NODE_ENV`. If so, it's also duplicated then.

.env.template
```diff
@@ -1,52 +1,37 @@
...
-TRANSLATION_TIMEOUT_MS="4000"
-SEARCH_SERVICE_URL="http://search-service:8000"
-SEARCH_REINDEX_TIMEOUT_MS="5000"
+POSTGRES_PORT="1111"
+POSTGRES_DB="db"
+POSTGRES_USER="user"
+POSTGRES_PASSWORD="password"
+
+# auth-service
+MONGODB_AUTH_URI=mongodb://localhost:11111/db
```
FPyMEHTAPIU, line 22: MONGODB_URI vs MONGODB_AUTH_URI mismatch. Template exposes only MONGODB_AUTH_URI ([.env.template:23](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)), but [docker-compose.yml:142](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1) sets MONGODB_URI directly (overriding env_file anyway). The template won't help the user debug if they override compose.

.env.template
FPyMEHTAPIU, line 1: Claude:

> Regressions from prior templates. The following were in the old per-service templates and now vanish without replacement: CORE_SERVICE_TIMEOUT_MS, TRANSLATION_API_URL, TRANSLATION_TIMEOUT_MS, SEARCH_SERVICE_URL, SEARCH_REINDEX_TIMEOUT_MS, GEMINI_GEN_MODEL, MONGO_ROOT_USERNAME/PASSWORD. Several are still read by code (translation.service.ts:24, searchIndex.service.ts:1, timeouts.ts:22) and only happen to work because compose hardcodes them.

FPyMEHTAPIU (changes requested):
The ports should be separated by name, since a single shared template can't give different services different ports.

FPyMEHTAPIU:
### Claude review:

> **Critical**
> - Dead PORT: entries in docker-compose.yml. Code was renamed on this branch to read PORT_GATEWAY / PORT_AUTH / PORT_CORE (api-gateway/src/index.ts:11, auth-service/index.ts:20, core-service/src/index.ts:11), but compose still sets the old PORT: 3000/3001/3002 ([docker-compose.yml:110,141,163](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)). Those overrides no longer exist in code — the whole port configuration is silently handed to env_file: .env, and the environment: override path (which was the point of listing them in compose) is gone. Either rename the compose keys to match or drop the rename; leaving both is misleading.
> - env_file: .env is required, but .env is gitignored with no bootstrap. Every service declares env_file: .env ([docker-compose.yml:4,25,41,71,98,132,154,185,218](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)). .env is in .gitignore. Nothing in Makefile or docs creates it from the template. Fresh clone → compose errors on missing file. Either make the file optional, provide a bootstrap step, or commit a .env.example that compose reads.

> **High**
> - PORT → PORT_GATEWAY/PORT_AUTH/PORT_CORE rename is a step backward. PORT is the standard every Node library, PaaS, and deploy tool expects. The underlying problem was "a single shared .env can't hold three different port values" — the fix should be per-service env files or per-service environment: blocks in compose, not renaming away from convention.
> - Interpolation without defaults removes compose's safety net. MONGODB_URI: ${MONGODB_AUTH_URI}, JWT_SECRET: ${JWT_SECRET}, TRANSLATION_API_URL: ${TRANSLATION_API_URL}, TRANSLATION_TIMEOUT_MS, SEARCH_SERVICE_URL, SEARCH_REINDEX_TIMEOUT_MS, CORE_SERVICE_TIMEOUT_MS, GEMINI_GEN_MODEL, MONGO_ROOT_USERNAME/PASSWORD ([docker-compose.yml:31,32,142,143,165–168,199,203,227](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)). None use ${VAR:-default} or ${VAR:?err}. If a key is missing from .env, compose silently substitutes empty string. Either require them (${VAR:?…}, like INTERNAL_SERVICE_TOKEN already does) or provide fallbacks.
> - Two independent keys that must stay in sync. MONGO_ROOT_USERNAME/MONGO_ROOT_PASSWORD initialize the Mongo container ([docker-compose.yml:31](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)), while MONGODB_AUTH_URI embeds creds inside the connection string ([.env.template:22](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)). The user has to keep them consistent by hand — a classic source of "worked for me" drift. Either derive the URI from the parts, or drop the parts and parse creds back out.
> - MONGODB_AUTH_URI vs MONGODB_URI naming. Auth code reads process.env.MONGODB_URI || process.env.MONGODB_AUTH_URI (auth_db_operations.ts:77); compose maps env_file's MONGODB_AUTH_URI → container's MONGODB_URI via interpolation ([docker-compose.yml:142](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)). Pick one name. The fallback in code suggests a prior migration that never finished.

> **Medium**
> - Template omits several keys the code actually reads. User-tunable env vars that don't appear in either the template or compose: GATEWAY_RESPONSE_TIMEOUT_MS, CORE_HEALTH_TIMEOUT_MS, AUTH_SERVICE_TIMEOUT_MS (api-gateway/src/utils/timeouts.ts:17,27,32), GEMINI_EMBED_MODEL, LLM_PROVIDER, OLLAMA_URL, SEARCH_DATABASE_URL, SEARCH_PRUNE_INTERVAL_SECONDS, TRANSLATION_MAX_TEXT_LENGTH / TRANSLATION_MAX_BATCH_ITEMS / TRANSLATION_MAX_TOTAL_CHARS. The whole point of a template is discoverability; if something is os.getenv'd, list it.
> - Template doesn't document DATABASE_URL vs POSTGRES_* precedence. Code picks DATABASE_URL if set, else assembles from POSTGRES_* (core-service/src/db/database.ts:16). The template lists both without a comment — previous per-service template had a "Preferred option / Alternative option" note that was dropped.
> - env_file: .env on pure-infra containers (traefik, auth-db, core-db, search-db). They don't read application vars. Dropping the directive is harmless and reduces noise. DOCKER_SOCKET_PATH reaches traefik through compose interpolation, not container env.
> - Stale header comment in [.env.template:2](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1): "Copy to .env in project root if needed". With env_file: declared on every service, it's no longer "if needed" — it's required.
> - Inconsistent quoting across the template: PORT_AUTH=1111 unquoted, PORT_GATEWAY="1111" quoted ([.env.template:6,24](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)). Purely cosmetic but odd for a reference file.

> **Low**
> - Missing trailing newline at EOF of .env.template.
> - Empty # general section header ([.env.template:39](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)).
> - Makefile inline DATABASE_URL ([Makefile:139](https://claude.ai/epitaxy/local_7b2ddebd-a4aa-49d7-b7b7-a6c7badb66d1)) — not a key/logic bug, just bypasses the consolidation the rest of the branch is doing. Test would be cleaner sourcing from the same .env.

el-kittiwake:
Everything I need is present in the .env.template

I've pushed relevant changes here.

I believe this commit will fix all issues raised by Claude in Nick's comment.

FPyMEHTAPIU (approved):

PR #264, opened 2026-04-21
User page

### LINKED ISSUE #226: Create user page + add UI for online/offline state

FPyMEHTAPIU:
## Summary
Adds a public user profile page at /user/:id and wires the supporting backend.

### Features

- User profile page (user.tsx): avatar, username, online/offline status, Follow/Unfollow, authored recipes, favorite recipes.
- Online status only shown to authenticated viewers.
- Follow/Unfollow with optimistic UI, guest-aware (opens auth modal, replays action after login).
- Favorites section visibility driven by backend 200/401/403 on /users/:id/favorites (mutual-follow gate enforced server-side).
- Self-profile redirect: visiting your own /user/:id redirects to /profile.
- Authored recipes grid now supports per-user mode via new userId prop on RecipesGrid (GET /users/:id/recipes).

### Backend

- GET /users/:id now returns status (via last_seen_at) and is_following (derived from requester).
- optionalAuth at api-gateway + extractUser at core-service so is_following survives refresh.
- Dropped is_followed_by — mutual-follow signal comes from /users/:id/favorites (200 vs 403).

### Layout / context

- Layout now reads id from /profile and exposes currentUserId via outlet context; refreshes on login, clears on 401/logout.

### i18n & a11y

- New userProfilePage block + aria-labels (profileAvatar, followUser, unfollowUser) in en/fi/ru.
- Loading/notFound → aria-live="polite", errors → aria-live="assertive".
- Follow button has contextual aria-label (Follow {name}) and aria-busy during requests.
- MainButton extended to forward aria-busy.

closes #226


SeelaSalorinta (approved):
Looks nice and works, tried testing with different accounts 💯 Here are some comments for you to think of:

1. The page never revalidates viewer-specific data after state changes. `/users/:id` and `/users/:id/favorites` are both viewer-dependent, but the load effect only depends on `id` / `isOwnProfile`, and follow/unfollow only updates local `isFollowing`. As a result, both logging in on this page and changing follow state can leave the follow button / gated favorites section stale until refresh.

2. If the user was already following before logging out, opening the page as a guest initializes `isFollowing = false`. After clicking Follow and logging in, the replay always sends `POST /follow`; if the server returns `409 already-followed`, the client treats that as failure and flips back to `false`, so the button can incorrectly show `Follow`.

3. I know this was changed intentionally, but I thought online status being limited to mutual followers was a useful privacy guard?

FPyMEHTAPIU:
Thank you @SeelaSalorinta! I'll fix the first two comments.
As for online status, I think we discussed that it should be displayed for all authorized users and only favorite recipes should be visible for mutual. But let's discuss it during the sync today 🙌

greyear (approved):
Smoke tested UI + checked BE stuff
Looks good

SeelaSalorinta (approved):
Excellent improvements, approved ! :)

PR #266, opened 2026-04-21
Jimik/recipe query search

gitenjoyer95:
Added recipe query search and removed the feature where Gemini gets the limit from users prompt. Now caps to 12 but fe can request with limits lower than that.

### DIFF
--- a/backend/services/api-gateway/src/__tests__/recipes.test.ts
+++ b/backend/services/api-gateway/src/__tests__/recipes.test.ts
@@ -50,6 +50,39 @@ describe("API Gateway - Recipes Routes", () => {
 		);
 	});
 
+	it("should forward recipe search query parameters to core-service", async () => {
+		fetchSpy.mockResolvedValue({
+			status: 200,
+			json: async () => ({
+				data: [{ id: 1, title: "Beef Tacos" }],
+				total_count: 1,
+				total_pages: 1,
+				page: 2,
+				per_page: 12,
+			}),
+		} as unknown as Response);
+
+		const response = await request(app).get("/recipes?q=beef&page=2&limit=12");
+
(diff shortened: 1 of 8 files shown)


el-kittiwake:
An empty query returns a string_too_short message.

A query with 501 characters returns a string_too_long message. 

A query such as `curl -s "http://localhost:3000/search/recipes?q=+"` returns message: q must not be blank

A nonsense string such as `curl -s "http://localhost:3000/search/recipes?q=yfasyufsfnsdlfhfsdlfdhflsfhsfnsufhsdjfsuf"` returns 12 recipes.

Searching for something non sequitur: `curl -s "http://localhost:3000/search/recipes?q=vainamoinen"` Returns 12, all with scores under 0.5. Should there be some threshold where it won't return the recipe if the score is too low?

A search for eggs: `curl -s "http://localhost:3000/search/recipes?q=eggs"` returns 12 recipes also. The first 5 contain eggs. Similar story with other simple keyword searches. Often the top ones are relevant, the rest not so much.

A more cryptic search: `curl -s "http://localhost:3000/search/recipes?q=find+me+cold+breakfast+recipes"` Returns overnight oats and mango smoothie bowl first, which I think makes sense. Changing to hot, returns shakshuka first, which I suppose could be breakfast?

Using the standard `curl -s "http://localhost:3000/search/recipes?q=give%20me%20the%20best%20beef%20recipe"` returns 12 recipes.

Reindexing works. Add a recipe, reindex it behaves.
&limit=1, 0, -1, 11, 12, 13 & string all behave as expected.

This all seems to be working pretty well.

el-kittiwake (approved):
Works fine. Doesn't look like it will break anything and doesn't actually seem to break anything.

PR #269, opened 2026-04-21
Profile page (supports editing profile, changing password, logout ,delete account)

### LINKED ISSUE #59: Create profile page

SeelaSalorinta:
Features
- users can update their username
- users can choose a preset avatar from avatar selection
- users can upload a custom profile photo
- uploaded photos show a preview before saving
- a default avatar is shown when no avatar is selected
- users can change their password from the profile page
- users can log out from the profile page
- users can initiate account deletion from the profile page
- success and error feedback is shown for profile actions and clears after five seconds
- after logout or account deletion, the user is redirected to the homepage
- translations updated and unused keys removed

Improvement after PR review:
- oversized profile photo uploads are blocked before save, with an error shown near the upload controls

close #59


frontend/app/assets/styles/profile.css
```diff
@@ -2,10 +2,210 @@
...
+	color: var(--text-secondary);
+	text-align: center;
+	margin-bottom: var(--space-m);
+}
+
+.profile-avatar-options-grid {
+	display: grid;
+	grid-template-columns: repeat(
+			auto-fit,
+			minmax(var(--preset-avatar-size), var(--preset-avatar-size))
```
FPyMEHTAPIU: minmax is redundant, since you use the same values
```suggestion
			var(--preset-avatar-size)
```

frontend/app/routes/profile.tsx
```diff
@@ -24,74 +38,88 @@ const ProfileResponseSchema = z.object({
...
 			})
 			.then((body: unknown | null) => {
-				if (ignore) {
-					return;
-				}
-
-				if (body === null) {
-					return;
-				}
+				if (ignore || body === null) return;
```
FPyMEHTAPIU, line 114: I liked the old formatting more 😅 
but it doesn't matter for now
```suggestion
				if (ignore || body === null) {
				    return;
				}
```

FPyMEHTAPIU (commented):
Hey! Nice work. The code is clean and understandable!
I tested avatars and other features and everything works nice! I left a couple comments and will leave Claude review, so maybe you can find some valuable points. I'd fix critical, hight and some medium/low issues. 🙌 

Claude review:

> 🔴 Critical
> 1. Preset avatars don't exist on disk (this seems fine to me, since you fetch them from the server – Nick)
> profile.tsx:41-44 generates paths /avatars/avatar-1.jpeg ... /avatars/avatar-8.jpeg, but frontend/public/ only contains favicon.ico — there is no public/avatars/ directory. The preset picker will render 8 broken s, and clicking one POSTs a non-existent path to PUT /profile. Either commit the 8 preset images under frontend/public/avatars/ or remove the picker until they exist.
> 
> 2. Trusting client-supplied userId for destructive ops
> profile.tsx:349-355 (DELETE /auth/delete) and profile.tsx:384-390 (POST /auth/logout) send { userId: authUser.id } in the body. Same pattern in ChangePasswordForm.tsx:71-76. If the backend honors this value instead of deriving it from the session cookie, an attacker can delete/log-out/change-password for any userId. Backend must ignore body-supplied userId and use the authenticated session. Worth verifying before merge.
> 
> 🟠 High
>
> 3. Wrong error string on delete/logout when authUser missing
> profile.tsx:335-337 and profile.tsx:367-369 surface t("profilePage.emailUnavailable") ("Email unavailable") when the user tries to delete their account or log out before /auth/me has loaded. The message is unrelated to the action. Either disable the buttons while isAuthUserLoading, or use a dedicated error key.
> 
> 4. id mismatch between profile and authUser
> ChangePasswordModal receives userId={profile.id} (profile.tsx:558) but delete/logout use authUser.id. They come from two different endpoints fetched independently. If they ever diverge (stale cache, races), behavior is silently inconsistent. Pick one source of truth.
> 
> 5. Hard page reload on logout/delete
> window.location.href = "/" in both handlers. This works but wipes SPA state and bypasses router. Fine as a shortcut for ensuring session cleanup, but worth standardizing — other auth flows in the codebase likely use router navigation + a state reset.
> 
> 🟡 Medium
>
> 6. Empty  fallback
> profile.tsx:492: src={resolveMediaUrl(avatarPath) ?? ""} in the preset grid. An empty src triggers a reload-of-current-page request in some browsers. Use defaultAvatar as the fallback here too, or skip rendering if unresolved.
> 
> 7. Avatar-only "save" sends avatar as a string path
> profile.tsx:260-264: formData.append("avatar", selectedAvatarPath) uses the same field name as the file upload but with a string value. The backend has to disambiguate file vs. string on a single multipart field. Cleaner to use distinct fields (avatarPreset vs avatarFile) — easier to validate and log.
> 
> 8. No client-side validation of file size/type beyond accept
> The accept attribute is advisory — users can pick anything via drag/drop or DevTools. A 50 MB file will be uploaded before the backend rejects it. Add a size guard in handleAvatarChange.
> 
> 9. Double-revoke of object URLs (benign but confusing)
> profile.tsx:195-201: the cleanup effect revokes avatarPreviewUrl on change, but handleAvatarChange and handlePresetAvatarSelect also revoke it manually before setting state. URL.revokeObjectURL is idempotent so this is harmless, but the intent is muddled. Pick one strategy.
> 
> 10. Handlers defined after early returns
> handleDeleteAccount and handleLogOut are declared after if (!profile) return ... (profile.tsx:334), while handleProfileSubmit is declared above. Works, but inconsistent; move them next to the other handlers.
> 
> 🟢 Low / Nits
>
> - Dead asset: frontend/app/assets/images/user-photo.jpg no longer has any importer — safe to delete.
> - AuthErrorResponseSchema duplication: defined locally in ChangePasswordForm.tsx:18-20; likely duplicates a schema used by the main auth modal. Consider extracting.
> - Magic string: "Password mismatch" (ChangePasswordForm.tsx:81) is a tight coupling to a backend error string — one typo on either side silently breaks the specific "current password incorrect" branch. Prefer matching on status code or a stable error code.
> - selectedAvatarPath not reset on profile refetch: if profile is reloaded while the user has an unsaved preset selection, the selection sticks. Minor.
> - usernameHint says "1-32" but required + maxLength=32 enforces that; there's no explicit min-length check beyond HTML required (which rejects empty, not spaces-only — mitigated by trim() in submit). OK.
> - Removed i18n keys (profilePage.title, userId, avatar, noAvatar, favoritesTitle, search.usersTab, createRecipe.createSuccess, ariaLabels.notifications) — verified via grep: no remaining references in app/. Clean removal.
> - Accessibility: fieldset/legend, aria-pressed, aria-live="polite", aria-hidden on decorative icons — all correct. Good work here.

SeelaSalorinta:
I fixed 3, 4, 5, 6, 9, 10.

I didn’t fully fix 2 because the remaining issue is backend-side: delete and change-password still use a userId request body as part of the current API contract. I verified the backend checks that userId against the authenticated JWT, and I removed userId from logout on the frontend since that endpoint doesn’t need it.

I didn’t change 7 because splitting preset avatar vs uploaded file into separate fields would require updating both frontend and backend request handling.

FPyMEHTAPIU (approved):
Looks really nice, awesome work! 😍 
My bad, I forgot to mention in the previous comment that we should probably duplicate a profile button to the footer as well, since it mirrors everything from the header except the search.
But Approving! 🙌

PR #270, opened 2026-04-21
Search filters

### LINKED ISSUE #228: Implement filters and connect in to the BE

FPyMEHTAPIU:
### Search page: filters, users tab, favorites wiring
Wires the search page's query/filter UI to the backend and ships a few small UI polishes.

**Search** params

- Forwards type, sort, ai, page, and multi-value filters (mealType, dishType, mainIngredient, cuisine) to /search/recipes.
- URL is the source of truth for all filter state; limit/sort fetch on change, filters only on Apply.

**Filter menu**

- New `SearchFilterMenu` dropdown (mirrors `SortMenu` positioning: absolute, outside-click + Escape close, doesn't take DOM space when closed).
- Four category multi-selects built on the existing `MultiSelectField`.
- A11y: <fieldset> with visually-hidden <legend>; submit-to-apply.

**Users tab**

- Added a "Users" pill next to Recipes. On the Users tab, Filter menu and AI toggle are hidden and `ai=0` is forced.

**Favorites**

- Favorite button on `RecipeCard` in results is wired via the shared `useRelationSet` composable (auth-gated, optimistic, rollback on failure, guest → login replay).

**Small UI fixes**

- Desktop language-dropdown chevron: now `--icons-default` (was dark-on-dark on the green header).
- Favorite button icon: `--icons-heart-active` outline, filled heart when active (previously overridden by `.icon-button svg`).

**Translations**

- `common.applyButton`, `searchPage.recipesTab`, `searchPage.usersTab` added for en/fi/ru.

closes #228


FPyMEHTAPIU:
I'd like to merge it after User Page, since that has some reusable composables and I'd like to avoid code duplication.

SeelaSalorinta:
Filters look really nice visually and feel smooth to use ✨

A couple things I noticed while testing the page - not sure which ones are in scope for this PR:

---

**1. Favorite icon outline regression**

The active favorite icon now has a white outline instead of `--borders-heart-active`. Was this intentional?

---

**2. Regression: follow/favorite replay after login**

This worked on main before, but now:

* guest clicks **Follow** or **Favorite**
* then logs in
* expected: button becomes active
* actual: stays inactive

Feels like a race where:

* auth triggers refetch of `/users/me/following` / `/users/me/favorites`
* that overwrites the optimistic replay state
* UI resets back to inactive

---

**3. Search in header**

Also noticed a UI issue: the search input stays visible in the header even after navigating away from Search to other pages like Home. 

---

**Some additional things (AI-assisted) you might want to look into:**

**4. Users tab is not wired end-to-end**

* still calls `/search/recipes`
* response validated as recipes
* state always set as `"recipes"`

So Users tab will either show recipes or break once user data is introduced.

---

**5. Query/filter params are frontend-only**

Frontend sends:

* `type`, `sort`, `page`, `ai`, filters

Backend currently only supports:

* `q`, `limit`

So filters, sort, AI toggle, pagination don’t actually affect results yet.

---

**6. Per-page options can break requests**

Frontend allows `24, 36, 48`
Backend caps at `12`

→ selecting higher values can return 422 and break the page

---

**7. Users tab URL state**

* `ai` can remain in URL when switching tabs
* recipe filters still included for Users

Could lead to inconsistent state when switching between tabs.

---

These are just my findings. I’ll let Jimi take the final approval 👍

FPyMEHTAPIU:
I fixed 1,2,3,4,7. Thank you for findings!
Other things are backend issues.

gitenjoyer95 (approved):

PR #279, opened 2026-04-22
Added notification fetch() to core when /auth/delete happens.

el-kittiwake:
Notify error, but make sure "user deleted" is returned.

Quick and dirty. Doesn't break existing functionality.


greyear (approved):
Looks good!

PR #280, opened 2026-04-22
Soft-delete in core

greyear:
Added POST /profile/delete to core-service to update new "is_deleted" field

Body:
{
    userId;
}


el-kittiwake (approved):
I think this looks like it won't screw us over later.

PR #282, opened 2026-04-22
Connect recipes and users pages to the be

### LINKED ISSUE #274: Move some tabs under auth protection
**Reporter**: @nikkxll

**Branch**: main

**Last commit**: Merge pull request https://github.com/greyear/transcendence/pull/266 from greyear/jimik/recipe-query-search (https://github.com/greyear/transcendence/commit/d57cee08d68c28e50ed186e55780b43f9e8d5c78)

**Side to fix**: FE

**Issue**: Remove My recipes and Saved recipes from the recipes view for non logged-in users and followers/following from users view

**How to reproduce**: open website -> go to all recipes / go to all users

**Devtools info / Docker logs**: -

**Possible solution**: Just put auth protection I guess

**Extra**:

Image
and
Image

FPyMEHTAPIU:
### Summary
Connects the Recipes and Users pages to the backend, replacing the previous static/mock data with real API integration.

**Recipes page**

- Tabs now drive live backend queries: All, My recipes (own authored), Saved (own favorites).
- New URL-scoped views: `?userId=…` shows a user's authored recipes, `?favoritesOf=…` shows another user's favorites (mutual-followers only). Both render with a back-to-profile link and a contextual title.
- Self-favorites links auto-redirect to the viewer's own profile.
- The profile page's "all recipes" and "all favorites" links deep-link into these scoped views.

**Users page**

- Tabs now drive live backend queries: All, Followers, Following (viewer-scoped).

**Shared improvements**

- Per-page dropdown (12 / 24 / 36 / 48) on both pages and search, persisted in the URL.
- Unified pagination row styling across recipes, users, and search.
- Auth-gated tabs show a sign-in prompt in the grid when a guest lands on them; the tab bar itself is hidden from guests to keep the UI coherent.
- Response parsing hardened with Zod schemas everywhere grids consume API data.

**Navigation**

- Added a Profile link to the footer for signed-in users.
- Removed the redundant "View all recipes" CTA from the home page.

**Backend**

- `/users/me/recipes` now returns picture_url so the My recipes tab can render cards with images.

closes #274 #271 #293


SeelaSalorinta (commented):
This is looking really nice overall. Visually it feels cleaner, hiding the auth-only tabs for guests makes the pages feel more functional, and the user-scoped recipe views are super cool.

A couple things to think about, and maybe not in this PR scope:
- The home page lost the big View all recipes CTA, and now the only recipes CTA is the small All link inside Top recipes. I’m not against removing the button, but I think we should decide which direction we want here:
        * either bring back the stronger View all recipes button for clearer navigation and remove the All link from Top recipes
        * or keep the smaller section link, but then maybe it should go to an actual top-recipes view instead of the generic /recipes page

- Related to that, I think Top recipes might not actually be top-rated anymore. The home page still passes sort="top", but the new /recipes pagination path seems to return before that sorting is applied. I already made a Kanban bug for this, so it can be checked separately.

- I still saw GET /users/:id/favorites 403 in the console on some profiles, but from the flow it seems expected for non-mutual followers, so I’m assuming that’s already known/tracked.

AI feedback:
- Sorting on the main recipes page looks page-local now. RecipesGrid.tsx fetches /recipes?page=...&per_page=... and then sorts only that returned slice on the client. Since the backend still paginates in created-time order, page 1 A-Z is not the global A-Z list, only that page sorted locally.

Overall though, this is a really solid improvement. Also thanx for removing my unused google translations haha!

FPyMEHTAPIU:
Nice findings, thank you!
I'd remove the big main button, but instead of sorting by rating let's have just generic /recipes. I don't believe top sorting actually works. 😅 

As for 403, I'd have a separate PR maybe to maybe print it as logs or ignore at all because subject says no errors, but it's not really an error, it's expected behavior. You can implement it as well, if you have time 😉

SeelaSalorinta (approved):
Nice!

el-kittiwake:
It is great to see all things coming together. A couple of things I'm curious about below, but they are likely just me being inexperienced and not really worth worrying about.

IMO, that big "View all recipes" looked a bit out of place. Looks much cleaner and simpler now.

---

Why does the front page fetch /recipes twice? Once for `https://localhost:8443/recipes?page=1&per_page=4`
and then shortly after for `https://localhost:8443/recipes?page=1&per_page=6` ?

Sometimes the front page shows 6 recipe cards and sometimes 4. I have not been able to find a pattern of actions that results in 4 consistently appearing.

---

When I go to recipes and then sort. It only seems to sort the recipes on that page. If I go to the next page, it is also only a sorted list of the recipes on that page.

So, I go to recipes, sort ascending, and I see for example: beef tacos, followed by chicken tikka masala.
I then go to the next page and see borscht and butter chicken.
Shouldn't those two have been sorted into the first page list?

FPyMEHTAPIU:
They will be properly sorted after the Backend pagination is connected. I'll check the double rendering in the evening

PR #286, opened 2026-04-22
stabilize Google button rendering and clear stale error state

### LINKED ISSUE #284: Stabilize Google sign-in button (see description)
## Problem

The Google sign-in button in the auth modal behaves inconsistently. The previous implementation mixed a custom-styled shell with the Google-rendered control, which led to layout/click inconsistencies and popup/button glitching when the modal state changed.

## Goal

Use a stable Google-rendered button in the auth modal without unnecessary rerenders.

### LINKED ISSUE #285: Reset stale validation state when switching auth modal mode (description)
## Problem

Validation errors from login/signup fields persist when switching between auth modes. For example, a password-related validation state can appear after changing modes even though the user has not interacted with that field in the new mode.

## Goal

Ensure login and signup start with clean field validation state when the mode changes.

SeelaSalorinta:
Fix two bugs in auth modal

What changed
1. BUG in google button clickability
- removed the custom Google button wrapper and now render Google’s official button directly
- stopped re-rendering the Google button when switching between login and signup
- switched the Google button text to the neutral Continue with Google variant
- constrained the Google button container to a width supported by Google
- localized the Google button based on the app’s active language
- added an OR divider between email/password auth and Google sign-in
- If google login is unavailable, Google -button and OR -divider are hidden 

2. BUG in error state in inputfields
- reset auth form inputs when switching between login and signup so validation state does not leak between modes (Previously, if the user focused or triggered validation on the password field in login mode and then switched to create account, the password field in the new mode could still appear red even though the user had not interacted with it there.)

close #284 
close #285

### DIFF
--- a/frontend/app/assets/styles/auth.css
+++ b/frontend/app/assets/styles/auth.css
@@ -17,41 +17,18 @@
 	gap: var(--space-m);
 }
 
-.main-button.inverted.auth-social-button {
-	background-color: var(--surface-card);
-	border: 1px solid var(--borders-action);
-	color: var(--text-primary);
-	height: auto;
-	min-height: 44px;
-	white-space: normal;
-	width: 100%;
-	pointer-events: none;
-}
-
-.main-button.inverted.auth-social-button[data-google-ready="true"]:disabled {
-	opacity: 1;
-}
-
-.auth-social-google {
-	stroke-width: 1;
-}
-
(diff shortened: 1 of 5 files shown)


greyear:
1) Now login button looks kinda strange. 

image

Could we have smth like this?

image

greyear (approved):

PR #291, opened 2026-04-23
 fixes the recipe translation flow end to end

### LINKED ISSUE #290: fix the recipe translation flow end to end

SeelaSalorinta:
It makes recipe fetches use the selected site language, makes recipe creation send the source language of the submitted text, and fixes background localization so translated values are actually saved after translation succeeds.

### Frontend
- Added `X-Language` to recipe list fetches in `RecipesGrid.tsx`
- Added `X-Language` to recipe detail fetches in `recipe.tsx`
- Added `X-Language` and `X-Source-Language` to recipe creation in `recipe-create.tsx`
- Added `X-Language` to user favorites fetches in user.tsx

### Backend
- Fixed the background localization update in `recipes.service.ts`
- The translation update could be skipped because PostgreSQL timestamp precision did not exactly match JavaScript `Date` precision
- The timestamp guard now compares at millisecond precision so successful translations are persisted

close #290

### DIFF
--- a/backend/services/core-service/src/services/recipes.service.ts
+++ b/backend/services/core-service/src/services/recipes.service.ts
@@ -296,14 +296,14 @@ const scheduleRecipeLocalization = (
 				`
 				UPDATE recipes
 				SET title = $1, description = $2, instructions = $3, updated_at = now()
-				WHERE id = $4 AND updated_at = $5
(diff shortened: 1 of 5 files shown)


SeelaSalorinta:
Good catch @gitenjoyer95 ,I updated `user.tsx` so `/users/:id/favorites` now also sends `X-Language`!

gitenjoyer95 (approved):

PR #292, opened 2026-04-23
added /auth/session to api gateway and use it for layout auth check in FE

### LINKED ISSUE #287: Chrome console on page load: GET https://localhost:8443//profile 401
Branch: main

Commit: 650a5d5

Issue: Every time I initially load a page I get the error. It only happens on reload or on opening the page in a new window/tab. It happens in any language. The page doesn't matter it seems to happen with any page that I can reach.

It seems to be trying to fetch the profile, but there is nobody logged in.

When logged in, this does not happen.

Error below, seems no way to get in plain text.

Image

greyear:
**BE changes**
GET /auth/session 
200 OK for every request
res.status(200).json({ authenticated: false });
or
res.status(200).json({ authenticated: true })

**FE changes**
call /auth/session to check whether the user is logged in
only if authenticated, call /auth/me to load the current user id
This avoids the guest-side /profile 401 in the browser console on initial load while keeping currentUserId available for existing UI logic

close #287  ?(hopefully)

### DIFF
--- a/backend/services/api-gateway/src/routes/auth.routes.ts
+++ b/backend/services/api-gateway/src/routes/auth.routes.ts
@@ -180,6 +180,24 @@ const patchChangePasswordHandler: RequestHandler = async (
 	}
 };
 
+const getSessionHandler: RequestHandler = async (
+	req: Request,
+	res: Response,
+	next: NextFunction,
+) => {
+	try {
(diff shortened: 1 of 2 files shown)


FPyMEHTAPIU (approved):
Looks nice to me 💪

PR #294, opened 2026-04-24
Add on-demand review translation

### LINKED ISSUE #296: Add on-demand translation for recipe reviews

gitenjoyer95:
## Backend
- Store review source language from `X-Source-Language`
- Add on-demand review translation endpoint
- Expose `source_language` in review responses

## Frontend
- Add Translate / Show original under reviews
- Hide Translate for reviews already in the current site language
- Keep translated text temporarily in React state so toggling between Translate/Show original does not call translation again; original `review.body` is unchanged

close #296

### DIFF
--- a/backend/infrastructure/postgres/core/01-init-core.sql
+++ b/backend/infrastructure/postgres/core/01-init-core.sql
@@ -337,6 +337,7 @@ CREATE TABLE "recipe_reviews" (
   "recipe_id" integer NOT NULL,
   "author_id" integer NULL,
   "body" text NOT NULL,
+  "source_locale" text NOT NULL DEFAULT 'en' CHECK ("source_locale" IN ('en', 'fi', 'ru')),
   "is_deleted" boolean NOT NULL DEFAULT false,
   "created_at" timestamptz DEFAULT now(),
   "updated_at" timestamptz DEFAULT now(),
--- a/backend/services/api-gateway/src/__tests__/recipes.test.ts
+++ b/backend/services/api-gateway/src/__tests__/recipes.test.ts
@@ -1113,37 +1113,86 @@ describe("API Gateway - Recipes Routes", () => {
 		expect(response.body).toEqual({ error: "Recipe not found" });
 	});
 
-	it("should proxy POST /recipes/:id/reviews/:reviewId/translate to core-service", async () => {
-		fetchSpy.mockResolvedValue({
+	it("should proxy GET /recipes/:id/reviews/:reviewId/translate to core-service", async () => {
+		fetchSpy.mockResolvedValueOnce({
+			ok: true,
 			status: 200,
-			json: async () => ({
-				data: {
-					review_id: 501,
-					recipe_id: 77,
-					source_language: "en",
-					target_language: "fi",
-					original_body: "Great",
-					translated_body: "Loistava",
-				},
-			}),
+			json: async () => ({ id: 42 }),
+		} as unknown as Response);
+
+		fetchSpy.mockResolvedValueOnce({
+			status: 200,
+			text: async () =>
+				JSON.stringify({
+					data: {
+						review_id: 501,
+						recipe_id: 77,
+						source_language: "en",
+						target_language: "fi",
+						original_body: "Great",
+						translated_body: "Loistava",
+					},
+				}),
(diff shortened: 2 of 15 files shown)


FPyMEHTAPIU (commented):
Looks and works nice! Great job!
I also asked Claude to review it, it has some points that are worth considering to fix:

> **High**
>
> - Unauthenticated translation endpoint is a cost/abuse vector. `recipes.routes.ts:455` wires `optionalAuth` on `/:id/reviews/:reviewId/translate`. The handler ultimately calls an external translation API per request and there's no rate-limit/throttle on it. Anyone scripting against this can run up bills or DoS the upstream. Recommend either `requireAuth` or a per-IP/per-user rate limit, and short-term in-memory caching of (`reviewId`, `targetLocale`) results since the underlying body rarely changes.
> - `translateTextToLocale` translates to all locales, not the requested one. `translation.service.ts:314` calls `localizeTextFromSource`, which fans out to `getTargetLocales(sourceLocale)` (every non-source locale) and then discards everything except `localized[targetLocale]`. With three locales, this is roughly 2× the upstream cost and latency it should be for the on-demand path. If the upstream API supports a single target, use it; otherwise add a small (text, source, target) cache to amortize the waste.

> **Medium**
>
> - Editing a review silently overwrites source_locale. `recipes.service.ts:1258` sets `source_locale = $2` on every update, defaulting to `DEFAULT_LOCALE` when the client omits `X-Source-Language`. There's no PUT-review caller in the frontend yet, but as soon as one is added without that header, a Finnish review becomes "en" on edit — and the test at `recipes.test.ts:2057` only happens to pass because it explicitly sets the header. Either keep the existing source_locale when the header is absent `(COALESCE($2, source_locale))` or require the header on update.
> - Gateway proxy crashes on non-JSON upstream responses. `recipes.routes.ts:81` does `await response.json()` unconditionally; if core-service returns an HTML error or empty body, the try catches it and returns 500 — but the message says "Failed to translate review" regardless of the upstream status (e.g. 502 from core becomes a 500 here). Consider response.text() + safe parse, or at least preserve the upstream status code.

> **Low**
>
> - `RecipeReview` type and `zod` schema disagree. `recipe.tsx:43` types `source_language: string`, while `RecipeReviewSchema` constrains it to `z.enum(["en","fi","ru"])`. Use `z.infer<typeof RecipeReviewSchema>` (or `SupportedLocale`) so the narrower type is reflected in usages.
> - `canTranslateReview` does string-equal compare on locales. `recipe.tsx:735` uses `review.source_language !== language`, where `language = i18n.resolvedLanguage ?? "en"` can be `"en-US"`. The button shows for an English review when the UI is en-US, then the backend (which truncates to 2 chars) returns the body unchanged. Normalize language to the first 2 chars before the comparison and before sending the header.
> - Translated-body cache is never invalidated when the underlying review changes. `translatedReviewBodies` is only cleared on full reload / language change. Edits or a "show original" after edit will display stale translations until the reviews list is refetched. Key the cache by (`reviewId`, `updated_at`) or clear the entry when a review is updated.
> - POST with no request body for a read-style operation — `recipes.routes.ts:993` is POST `.../translate` with no payload. GET would be more conventional, cacheable by intermediaries, and trivially idempotent. Not blocking, but worth a thought.
> - Default 'en' for historical rows is a silent assumption. The `01-init-core.sql` change defaults existing rows to English — fine for fresh dev DBs, but if any environment already has non-English reviews seeded, they'll mis-translate going forward. Confirm there's no production data to migrate or document the assumption.

> **Nit**
>
> - Test at `recipes.test.ts:1825` translates fi → fi, which exercises only the early-return path in `translateTextToLocale`. Add a case where source ≠ target so the actual upstream call path has coverage.
> - New i18n keys are added in three languages — good — but no test that the translateReview button surfaces the error string when status is non-OK.
> - Summary
> - Functionally the feature works and tests are present, but there are two items I'd resolve before merge: gating/rate-limiting the unauthenticated translation endpoint, and fixing `translateTextToLocale` so it doesn't translate to locales it then throws away. The `source_locale` overwrite-on-update is a latent bug that becomes real the moment a review-edit UI lands.

FPyMEHTAPIU (approved):
Nice fixes!

PR #295, opened 2026-04-24
Fix bug:  initialize Google Identity only once

### LINKED ISSUE #289: Chrome console when opening login dialog: [GSI_LOGGER]: google.accounts.id.initialize() is called multiple times.
Branch: main

Commit: https://github.com/greyear/transcendence/commit/650a5d577ccbbfb0f98e3f4345f00fcd176845d6

Issue: When opening a login dialogue (only on second and subsequent openings) the below warning appears.
This happens: clicking Sign In, trying to do something (rate, follow, etc.) that requires being logged in.

It only happens on the second time and subsequent times of opening a login dialogue. Reloading the page will reset this "count".

Image

SeelaSalorinta:
Fixes repeated Google Identity Services initialization when reopening the auth modal.

close #289


el-kittiwake (approved):
Fixes the mentioned bug and does not appear to break anything.

PR #306, opened 2026-04-25
Reindexing at startup

gitenjoyer95:
Added reindexing recipes into search db on starup and one line change to handle >12 inputs as a limit on AI search better


FPyMEHTAPIU (commented):
Seems good, but I can see translation related commits. I assume, it should be merged after the translation reviews.
Claude review as well (doesn't seem like a blocker):

> **Medium**
> - count_search_documents ignores embedding state. main.py:367-373 counts every row in recipe_search_docs, but search_recipe_documents only returns rows where embedding IS NOT NULL (main.py:488). If a prior run inserted rows but failed before embedding was set, startup will skip reindex while searches return nothing. Filter with WHERE embedding IS NOT NULL.
> - Reindex runs in a daemon thread. main.py:662-668 spawns the worker as daemon=True, so a container shutdown mid-bootstrap kills the thread abruptly — the job stays running in memory and the DB is partially populated, but the in-process worker lock is gone with the process anyway. Acceptable since UPSERT is idempotent and the next start will re-check, but worth a non-daemon thread + a shutdown handler if you want clean shutdowns.

> **Low**
> - Deprecated @app.on_event("startup"). main.py:662 — modern FastAPI prefers the lifespan context manager. Works today, emits deprecation warnings on newer FastAPI.
> - No client-facing signal during bootstrap. While the daemon thread is running, /search/recipes returns empty data with no hint that the index is still warming up. Consider surfacing the active job status in /health or a 503 fast-path.

PR #307, opened 2026-04-25
Wcag fix

### LINKED ISSUE #224: Check WCAG 2.1 compliance and fix issues

FPyMEHTAPIU:
### Accessibility fixes (WCAG 2.1 AA)
Patterns applied across the frontend, not per-element:

- Region landmarks earn their place. `<section>` is reserved for thematic groups that own a heading; pure layout wrappers are `<div>`. Cuts landmark noise — recipe page went from 5 regions to 3.
- Composite widgets get a single accessible name. Multi-node displays (e.g. rating: number + icon + count) wrap with `role="img"` + translated `aria-label`, inner nodes `aria-hidden="true"`. Announced as one unit instead of three fragments.
- Native semantics over ARIA shims. Status messages use `<output>` instead of `<div role="status">` (also satisfies Biome useSemanticElements).
- Decorative icons hidden. Inline icons inside labeled widgets are `aria-hidden="true"` so the name comes from one source.
- i18n-aware ARIA text. Every new aria-label has en/fi/ru entries — no English-only labels.
- SCs touched: 1.3.1 Info and Relationships, 4.1.2 Name, Role, Value.

closes #224


greyear:
### 1. `role="toolbar"` on `FilterList` is the wrong role

**File:** `app/components/FilterList.tsx`

`role="toolbar"` carries an implicit contract that arrow keys navigate between its items. That keyboard behaviour is not implemented. A screen reader user will reach the toolbar, expect arrow key navigation, and find none. Use `role="group"` instead — which is what the audit actually recommended:

```tsx
<div className="filter-list" role="group" aria-label={ariaLabel}>
```

---

### 2. `<button role="option">` in `SortMenu` is invalid ARIA

**File:** `app/components/SortMenu.tsx`, rendered `<button type="button" role="option" ...>`

ARIA spec prohibits placing `role="option"` on an element that already has a native interactive role (`<button>`). Some screen readers will announce "button" plus "option", or will treat it inconsistently. Replace the `<button>` wrappers with non-interactive elements:

```tsx
<div
  key={option.value}
  role="option"
  aria-selected={isSelected}
  tabIndex={index === activeIndex ? 0 : -1}
  ref={(el) => { optionRefs.current[index] = el; }}
  className={...}
  onClick={() => handleSelect(option.value)}
  onKeyDown={(e) => {
    if (e.key === 'Enter' || e.key === ' ') {
      e.preventDefault();
      handleSelect(option.value);
    }
  }}
  onFocus={() => setActiveIndex(index)}
>
  {option.label}
</div>
```

---

### 3. Mobile header overlay: background content remains reachable by Tab

**File:** `app/layouts/Header.tsx`

Focus management (move in on open, return to toggle on close, Escape handling) is implemented correctly. But the overlay has no `role="dialog"` / `aria-modal="true"`, and the page behind it is not marked `inert`. A keyboard user who Shift-Tabs past the first overlay item will fall into the background navigation. Add either approach:

**Option A (minimal):** Add `role="dialog" aria-modal="true"` to the overlay div — most screen readers will then restrict virtual cursor to the overlay.

```tsx
<div
  ref={overlayRef}
  id={overlayId}
  role="dialog"
  aria-modal="true"
  aria-label={t("ariaLabels.toggleMenu")}
  className="header-menu-overlay"
>
```

**Option B (belt-and-suspenders):** Also set `inert` on the `header-top-row` sibling while the overlay is open.

---

### 4. `LanguageSelector` menu has no arrow-key navigation

**File:** `app/components/LanguageSelector.tsx`

Escape closes the dropdown and the trigger has correct `role="menu"` / `aria-haspopup="menu"` / `aria-controls`. But once focus enters the menu, ↑/↓/Home/End do nothing. The ARIA menu pattern requires roving focus. Add a `keydown` handler on the `<div role="menu">` similar to what was done for `SortMenu`.

---

### 5. `StarSolid` in `RecipeCard` card footer missing `aria-hidden`

**File:** `app/components/cards/RecipeCard.tsx`, line 47

```tsx
// Current — missing aria-hidden
<StarSolid />

// Fix
<StarSolid aria-hidden="true" />
```

Every other star/icon in the diff got this treatment (recipe detail page, pagination arrows, etc.), but the card footer star was missed. A screen reader will announce it as an unlabelled graphic.

---

### 6. `skip-link` transition not guarded by `prefers-reduced-motion`

**File:** `app/index.css`, line 45

Every other transition in this PR is conditionally applied via `@media (prefers-reduced-motion: no-preference)`, but the skip link's `transition: transform 0.15s ease` is unconditional. Wrap it:

```css
@media (prefers-reduced-motion: no-preference) {
  .skip-link {
    transition: transform 0.15s ease;
  }
}
```

FPyMEHTAPIU:
Fixed 1,4,5,6

greyear (approved):
All three issues are now fixed. Approved!

PR #308, opened 2026-04-25
Translations back

greyear:
Localized recipe dictionaries and recipe content flow across the stack: migrated ingredients, units, and categories to JSONB, added translation-service handling for recipe text, fixed seed data and locale-aware API responses, and updated the recipe UI to display localized units


backend/services/core-service/src/services/translation.service.ts
```diff
@@ -166,7 +166,9 @@ const requestTranslations = async (
 	if (!TRANSLATION_API_URL) {
 		return null;
 	}
-
+	console.log(
```
el-kittiwake, line 169: Should we have console.log?

backend/services/core-service/src/services/translation.service.ts
```diff
@@ -183,6 +185,7 @@ const requestTranslations = async (
 			}),
 			signal: timeoutSignal(),
 		});
+		console.log(`Translation API responded with status ${response.status}`);
```
el-kittiwake, line 188: console.log() again

backend/services/core-service/src/services/translation.service.ts
```diff
@@ -193,6 +196,7 @@ const requestTranslations = async (
 		}
 
 		const data = (await response.json()) as TranslationApiResponse;
+		console.log("Translation API response data:", data);
```
el-kittiwake, line 199: console.log()

el-kittiwake (approved):
As far as i can see it all works and doesn't break anything.

I would say the changes look consistent and understandable, so I am confident that I have not missed anything.

I asked copilot to have a look too, just to be sure.

backend/services/core-service/src/services/translation.service.ts
greyear: This is gonna be shown only in container, not in browser

backend/services/core-service/src/services/translation.service.ts
greyear: Same

backend/services/core-service/src/services/translation.service.ts
greyear: Same

PR #312, opened 2026-04-25
Add recipe delete flow and metadata display

### LINKED ISSUE #297: Show full recipe metadata on recipe details page

SeelaSalorinta:
## Summary

- Adds recipe detail metadata for created/updated dates, categories, servings, cook time, and spiciness.
- Adds recipe owner delete flow with a confirmation modal, disabled deleting state, error handling, and redirect after success.
- Hides rating and review actions from the recipe author.
- Makes review and created/updated dates locale-aware.
- Makes auth modal errors clear automatically after 5 seconds.
- Removes unused auth success message UI because the auth modal closes immediately on success.

close #297


greyear:
Hey Seela! Overall changes look very neat and I like the style of newly created fields on FE. Even though updated_at field won't be changing even and will be probably always the same as created_at.
I noticed a little inconsistency here cause for end and fi words start with capital letters.
image
Ingredients and units are just perfect now in all languages.

Also AI noticed these things, Idk if this is critical or not, but take a look pls:
1) In AuthForm, errors auto-dismiss after 5 seconds via a useEffect timeout — nice UX touch. But deleteRecipeDeletionError has no equivalent: once it appears, it stays there until the user does something else. Not a blocker, but it's inconsistent with the pattern established elsewhere in this same PR.
2) In deleteRecipe, the happy path looks like this:
```
navigate("/recipes")  ← component starts unmounting
...
finally {
  setDeletingRecipe(false)        ← called on unmounted component
  setIsDeleteRecipeModalOpen(false) ← same
}
```
finally always runs, even after a successful navigate. By the time those state setters fire, the component may already be gone. React will likely log a warning like "Can't perform a React state update on an unmounted component". The fix is simple — just return early after navigate, before the finally block runs those setters, or restructure so navigation only happens after finally.

greyear (approved):

Issue #301, opened 2026-04-24
Filters not working for plain recipe page


el-kittiwake:
Branch: main

Commit: ad999b0

Browser: I am using the latest Firefox on Hive machines.

Issue: Trying to filter recipes on the recipe page when I have not performed a search does not work as intended I believe.

Clicking the "Filter" button does nothing.

PR #315, opened 2026-04-25
Pagination fix

### LINKED ISSUE #302: Filtering search results doesn't do what it is supposed to
Branch: main

Commit: ad999b0

Issue: When I perform a search (with non-seed-data results) and attempt to filter using a time/dish/ingredient/cuisine that was not marked on those dishes, the dishes remain in the list of results anyway.

Gubbins

Image

With these all set or individually set, none of the non-seed recipes should be returned.

Image

greyear:
- Fixed users grid
- Fixed users filters to apply correctly
- Fixed recipes filters to apply at all

closes #302 #301 #236


SeelaSalorinta (approved):
Hey nice job! Pagination works nicely! Here are some Ai comments for you to read

Two clear regressions in this PR:

1. Followers/following pagination regressed.

`UsersGrid` now renders `userList` directly and sends `page`, `per_page`, and `sort` to every users tab. But `/users/me/followers` and `/users/me/following` still return all rows as `{ data, count }`, not a paginated response. On `main`, these tabs were sorted and sliced client-side, so pages worked for all-at-once follower/following responses.

2. Username search relevance regressed.

In `getAllUsers`, prefix-match ranking was moved after `${sortExpr}`. On `main`, `CASE WHEN u.username ILIKE $2 THEN 0 ELSE 1 END` ran before the sort expression, so prefix matches ranked first. With the PR default `u.id ASC`, ID ordering can now beat relevance.

Not necessarily introduced here, but this PR makes it more visible: Finnish recipe search for terms like `kana` still does not work reliably. The search page does not send `X-Language`, and the default AI search path indexes English-first recipe documents. Since this PR makes AI search the default, Finnish searches are more likely to hit the English-biased path.

greyear:
@SeelaSalorinta we discussed with Nick, decided to remove sort and filter from /users view at all

PR #316, opened 2026-04-25
Add custom 404 route with proper status response

### LINKED ISSUE #304: Make 404 page

SeelaSalorinta:
## Summary
- Added a catch-all frontend route for bad URLs that do not match any app route, like `/bad-url`
- Rendered the shared `NotFoundView` for unmatched paths so users see the custom 404 page
- Set the catch-all route loader response status to `404` so missing frontend pages return the correct HTTP status
- Reused the same not found view for valid routes with missing data, like `/recipes/999`, when the API returns `404`

close #304


frontend/app/components/NotFoundView.tsx
```diff
@@ -0,0 +1,32 @@
...
+			<div className="not-found-page-content">
+				<h1 id="not-found-title" className="not-found-page-title">
+					<span>{t("notFoundPage.titleLine1")}</span>
+					<span>{t("notFoundPage.titleLine2")}</span>
+				</h1>
+				<MainButton
+					to="/"
+					variant="primary"
+					className="not-found-page-button"
+					aria-label={t("notFoundPage.backHome")}
```
greyear: just
```
<MainButton
  to="/"
  variant="primary"
  className="not-found-page-button"
>
  {t("notFoundPage.backHome")}
</MainButton>
```

greyear (changes requested):
One minor thing

greyear (approved):
Thanks!

PR #317, opened 2026-04-25
Add startup search reindex bootstrap

gitenjoyer95:
Created new branch and applied reindexing logic manually to avoid merge chaos

### DIFF
--- a/backend/services/search-service/app/main.py
+++ b/backend/services/search-service/app/main.py
@@ -364,6 +364,15 @@ def update_reindex_job(job_id: str, **updates: Any) -> None:
             job.update(updates)
 
 
+def count_search_documents() -> int:
+    with psycopg.connect(SEARCH_DATABASE_URL) as connection:
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT COUNT(*) FROM recipe_search_docs")
+            row = cursor.fetchone()
+
+    return int(row[0]) if row else 0
+
+
 def reindex_error_message(error: Exception) -> str:
     if isinstance(error, HTTPException):
         return str(error.detail)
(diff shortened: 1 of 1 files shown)


FPyMEHTAPIU (approved):
Looks nice to me. Read Claude's review as well, and decide if something is worth fixing.

> ## Concerns
> 
> ### 1. Deprecated `@app.on_event("startup")`
> FastAPI replaced `on_event` with lifespan context managers since v0.93. Still functional but generates deprecation warnings and diverges from recommended convention.
> 
> ### 2. `row` may be unbound in `count_search_documents`
> `row` is assigned inside the `with connection.cursor()` block and referenced after it closes. If `fetchone()` raises, `row` is unbound and the `return` statement throws `UnboundLocalError`. This is caught by the caller's `except Exception`, but it's a fragile pattern. Fix: initialize `row = None` before the cursor block.
> 
> ### 3. Multi-worker race condition
> `_reindex_jobs` and `_reindex_worker_lock` are in-memory and per-process. If the app runs with multiple Gunicorn/uvicorn workers, each worker independently checks `count_search_documents()` and may trigger a simultaneous full reindex. The per-process lock provides no protection across workers.
> 
> ### 4. No test coverage
> `startup_reindex_if_index_is_empty` and `count_search_documents` have no tests.
> 
> ## Verdict
> 
> **Safe to merge for single-worker deployments.** Exception handling is solid; the startup thread won't destabilize the service.
> 
> **Block if running multiple workers** — the race condition would cause redundant simultaneous full reindexes on startup.
> 
> **Recommended before merge:**
> - Initialize `row = None` before the cursor block in `count_search_documents` (low-risk, one line)
> - Confirm deployment runs a single worker, or add a distributed lock / DB-level guard
> 
> **Nice to have:**
> - Migrate to a lifespan handler to clear the deprecation warning
> - Add at least one unit test for the empty-index path

PR #319, opened 2026-04-25
Css migration

### LINKED ISSUE #275: implement some css framework part and migrate to scss.

FPyMEHTAPIU:
## Summary

- Migrated all 39 CSS files to SCSS (`sass` added as dev dependency, Vite picks it up with no config changes)
- Introduced `_variables.scss` (breakpoint constants) and `_mixins.scss` with 9 reusable mixins: `flex-center`, `flex-col`, `flex-col-center`, `interactive-brightness`, `button-disabled`, `surface-card`, `avatar-image`, `status-text`, `breakpoint`
- Refactored 14 component files to use `@use`, mixins, nesting, and `@include breakpoint` — eliminating repeated patterns like the identical `.auth-card`/`.confirmation-card` blocks, 4 circle avatar elements, and 3 status text elements

## Why

Plain CSS did not satisfy the project requirement for a modern CSS framework or styling solution. SCSS provides real preprocessing features (variables, mixins, nesting, modules) that plain CSS cannot replicate.

closes #275


SeelaSalorinta (commented):
Niceee!

Findings

P2: Reduced-motion behavior regressed in the shared mixin. _mixins.scss (line 20) now emits transition: filter 0.2s ease unconditionally, while the previous button/menu styles guarded these transitions with @media (prefers-reduced-motion: no-preference). cooksRow.scss (line 19) has the same issue for the fade transition.

P2: The migration is incomplete. NotFoundView.tsx (line 4) still imports not-found.css, and categoryFilterMenu.css remains in frontend/app/assets/styles as a dead duplicate of searchFilterMenu.scss. That means the branch does not actually satisfy “all CSS files migrated to SCSS.”

P2: Disabled pagination styling regressed. pagination.scss (line 66) now uses opacity: 0.3 for disabled links/buttons instead of the previous explicit text/icon colors. That lowers contrast for the whole control and also conflicts with the WCAG audit note that specifically calls out avoiding opacity-based disabled pagination styling.

P3: The danger button lost its custom border color. mainButton.scss (line 37) only sets the error background now; the previous CSS also set border-color: var(--borders-input), so danger buttons now keep the default action border.

ALso does mobile look different everything is smaller or should it be like that , shows six recipes not four home page, looks different than before, maybe its the new normal

Image

SeelaSalorinta (approved):
Excellent! :)

PR #322, opened 2026-04-25
Fix ingredient fields in recipe create

SeelaSalorinta:
- Make ingredient/unit field sizing target the actual rendered input/select wrapper classes
- Widen the unit select from 130px to 180px! More space to make Yksikkö show in finnish!

## Notes
This prevents shared `SelectField` / `InputField` `width: 100%` styles from overriding the recipe row layout and causing the fields to stretch incorrectly


FPyMEHTAPIU (approved):
Nice fix, thank you 🫶

Issue #321, opened 2026-04-25
Recipe cards not translated in search results


el-kittiwake:
Branch: main

Commit: f58e4e4

Browser: I am using the latest Chrome on Hive machines.

Issue: The search result recipe cards are not translated to chosen language.

<img width="2134" height="1795" alt="Image" src="https://github.com/user-attachments/assets/230996dd-1cb6-4483-8170-aa32f1911cb6" />

<img width="2134" height="1795" alt="Image" src="https://github.com/user-attachments/assets/e11cec8b-f38f-4f80-8a3c-a3287a6fab5c" />

PR #323, opened 2026-04-25
fixing translations bug with i18next

greyear:
Fixing issue:
Recipe cards not translated in search results #321


el-kittiwake (approved):
Search behaviour is better now. No longer shows the wrong language recipe cards when changing language.

Issue #313, opened 2026-04-25
Recipe creation page, serving number not validated in the field - 400 error on attempted create


el-kittiwake:
Issue: Entering a large serving size in recipe creation does not inform the user anything until creation button is pressed. Maximum allowable number is too large anyway. Unless we're Jesus or the Roman Army.

<img width="1988" height="296" alt="Image" src="https://github.com/user-attachments/assets/76758217-fa48-450b-89f5-b88fe72da2e8" />

Create fails so some validation is happening somewhere

<img width="1988" height="296" alt="Image" src="https://github.com/user-attachments/assets/dd27b08a-36f4-4bc1-a7b8-998f8660b484" />

Max value that gets past that error: 999999999999999

Attempting to create a recipe with such a large value results in 400.

<img width="2037" height="262" alt="Image" src="https://github.com/user-attachments/assets/aa4ab484-08b0-426a-b63c-7fbeadd0acac" />

<img width="843" height="792" alt="Image" src="https://github.com/user-attachments/assets/71d8c3aa-09fe-4c76-a9e0-2c5b67ec3e26" />

Maximum successful number: 999999999

PR #324, opened 2026-04-25
fix for servings limit

greyear:
Fix for:
Recipe creation page, serving number not validated in the field - 400 error on attempted create #313


el-kittiwake (approved):
Fixes the bug in question.

PR #327, opened 2026-04-25
AI search return pictures

gitenjoyer95:
AI search now returns pictures

### DIFF
--- a/backend/infrastructure/postgres/search/01-init-search.sql
+++ b/backend/infrastructure/postgres/search/01-init-search.sql
@@ -6,10 +6,14 @@ CREATE TABLE IF NOT EXISTS recipe_search_docs (
   description text,
   instructions text NOT NULL,
   searchable_text text NOT NULL,
+  picture_url text,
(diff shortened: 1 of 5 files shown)


SeelaSalorinta (commented):
AI comments

Blocking

backend/services/search-service/app/main.py

/search/recipes now returns row["picture_url"], but search_recipe_documents() does not select picture_url from recipe_search_docs. Any AI search with matching rows will crash with a missing key instead of returning results.

Suggested fix: include picture_url in the SELECT and returned row mapping inside search_recipe_documents().

Important

backend/services/core-service/src/services/recipes.service.ts

Picture uploads for already-published recipes do not trigger a search reindex, and updateRecipePicture() also does not update recipes.updated_at. That means AI search can keep returning stale or null picture_url values until a full reindex happens.

Suggested fix: after successfully updating recipe_media, call scheduleRecipeSearchReindex(recipeId) when the recipe status is published.

SeelaSalorinta (approved):
Nice fix

PR #330, opened 2026-04-25
Deletion fixes + db retries

greyear:
- Retries for db added to makefile
- Deletion changes, some FE fixes for deleted recipes


SeelaSalorinta (commented):
Findings

RecipeCard.tsx (line 3) fails npm run check:frontend because imports are not sorted. Biome wants useTranslation before Link, so this will block the frontend check.

RecipeCard.tsx (line 61) still renders the favorite button for archived recipe cards. RecipesGrid passes isArchived at RecipesGrid.tsx (line 322), but the card only uses it for the badge. Backend favorite mutations require published recipes, so clicking the heart on an archived recipe will optimistically toggle and then silently roll back with a 404. The recipe detail page already hides favorites for archived recipes at recipe.tsx (line 874); cards should match that behavior or disable the action.

SeelaSalorinta (approved):
Nice job!

PR #331, opened 2026-04-25
less filler recepies that arent really similar to search

gitenjoyer95:
ai seaerch Filters the results more strictly now


greyear (approved):
Looks good, no major stuff

PR #332, opened 2026-04-25
Seed categories

greyear:
1) added reviews for seed recipes
2) cook_time added for seed recipes
3) categories are now translated


el-kittiwake (approved):
Seed data fits in fine. No replicable problems. That one issue may forever remain a mystery.

PR #341, opened 2026-04-26
Add backend to traefik and mask it via middleware.

FPyMEHTAPIU:
Change frontend cors service port to `5173/api`

### DIFF
--- a/traefik/dynamic.yml
+++ b/traefik/dynamic.yml
@@ -6,12 +6,27 @@ http:
         - websecure
       tls: {}
       service: gateway
+    gateway-on-frontend:
+      rule: "(Host(`localhost`) || Host(`127.0.0.1`)) && PathPrefix(`/api`)"
+      entryPoints:
+        - websecure-frontend
+      tls: {}
+      service: gateway
+      middlewares:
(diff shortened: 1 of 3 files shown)


traefik/dynamic.yml
```diff
@@ -6,13 +6,28 @@ http:
...
       service: gateway
+    gateway-on-frontend:
+      rule: "(Host(`localhost`) || Host(`127.0.0.1`)) && PathPrefix(`/api`)"
+      entryPoints:
+        - websecure-frontend
+      tls: {}
+      service: gateway
+      middlewares:
+        - api-strip-prefix
+      priority: 100
```
greyear, line 17: Missing explicit priority on frontend-https Traefik's auto-priority is based on rule length, so the longer (Host(...)) && PathPrefix('/api') rule would win anyway. But since you've added explicit priority: 100 to one router, leaving frontend-https without one creates an asymmetry that could trip up someone later. Consider adding priority: 10 (or any value below 100) to frontend-https to make the intent self-documenting.

greyear (changes requested):

greyear (approved):

PR #343, opened 2026-04-26
Limit fixes etc. to fe

SeelaSalorinta:
## Summary
- Add title length validation and improve wrapping for long recipe text
- Add password length validation for login and change-password flows
- Improve disabled pagination styling
- Add frontend recipe instruction limits:
  - 50 instruction steps max
  - 1000 characters per step
  - 10000 characters total
- Disable the Add step button when the instruction step limit is reached
- Add localized validation messages for new recipe limits


FPyMEHTAPIU (approved):
Looks good and works nice! 🚀 

Only one thing noticed, that there's no error message under the title field and no counter. Otherwise, awesome! Approving ✨

